PR #1, opened 2019-06-21
Set the max age for cookies

lolametro:
Set the max age for cookies to 30 days (has to be given in millisecs).
I couldn't test the change since I don't have the necessary Discord API keys.


Sendouc:
Making the log-in persistent needs some other kind of changes to work so closing this for now.

PR #12, opened 2020-06-28
Adds first few pt-BR pages to check if the huge sentences are breakin…

guigas7:
…g the UI.

I'll format the json for the final version

Signed-off-by: Guigas <jimhorton7@outlook.com>


Sendouc:
Can you please format it already? You can do so easily with VS Code + Prettier https://github.com/prettier/prettier-vscode when you have it installed simply go to the document and press CTRL+Shift+P and look for "Format document"

guigas7:
actually i'll check it with inspect element, since each string will have it's own ideal size, i think i'll send the final version this tuesday.

PR #18, opened 2020-07-01
Translations for Traditional Chinese

telgniw:
## What is translated in this PR
- sub weapons
- special weapons
- hero weapons
- stages (non-splatfest)
- modes

## What might be translated in this PR
- main weapons: might need more discussion because our naming is really messy IMO

## Notes
- Two scripts are added for my own convenience when doing translations. I can drop the commit if there's better solutions for command-line translators to convert between utf8 and unicode-escaped strings.


Sendouc:
Sounds good. What do you mean with "main weapons: might need more discussion because our naming is really messy IMO"?

telgniw:
> What do you mean with "main weapons: might need more discussion because our naming is really messy IMO"?

Hmm... for sub weapons I might know two names used equally frequent and I can make it like "torpedo (soy sauce)". It won't look messy because both names take only 2 (w-)chars.

But for main weapons like "E-liter 4K" it might be "no-brand 4K" or "4K with rain", or just "4K", or even "no-scope 4K" maybe? I definitely don't want to make it "no-brand no-scope 4K with rain". :joy:

Another not-so-messy example is the splattershots. I usually call them like "splattershot with burst bomb", "splattershot with inkjet", and "the black-and-white splattershot". But when it comes to translations I'm suddenly not sure if they should have such inconsistent and asymmetric names (one named by sub, one by SP, and one by color). I tend to not mention the brand for weapons since almost no one knows the brands.

I personally prefer to finish this PR before main weapon names are all filled.

telgniw:
Sorry to left it here for a while. I just re-checked and pushed the commit for the 5 listed groups.

Sendouc:
Thank you!

PR #44, opened 2020-07-28
Improve German localization

lolametro:
Various minor changes for German localization, mainly:
- Bild -> Frame (otherwise hardly understandable without context)
- Clarified Quick Respawn effect
- Uppercased first words of a sentence
- Teamsuch-Post -> Beitrag zur Teamsuche (still not optimal but a bit less colloquial)


Sendouc:
Thanks!

PR #235, opened 2020-12-30
[pt_BR translation] Adds last fields i forgot + minor fixes

guigas7:
2020 plz end already


guigas7:
I messed one thing, just a sec

PR #313, opened 2021-03-28
Update development setup instructions

rofreg:
Hi there! I just got my local development environment up and running (🎉), and I thought it would be useful to update the setup instructions to fix a few issues that I encountered.

One important place where I could use some debugging help: `dotenv` and `npm run prebuild` didn't quite want to work together for some reason! In my experiments, `preBuild.ts` could successfully load ENV vars from `.env`, but not from `.env.local`. I fixed this by manually including the ENV vars in `prisma/.env` as well, but I'm not sure that's the best strategy – the local setup that you're already using in your own development may be better. Where are you currently setting `DATABASE_URL` in your various `.env` files?


Sendouc:
🎉 

First of all thank you for the feedback! I haven't had too many people try to run the project locally so I'm sure there is a lot that can be improved. I really appreciate the chance to improve this process.

Currently locally I'm actually using the same approach: (it is detailed here: https://github.com/Sendouc/sendou.ink#access-rest-of-the-pages) but you are right that it would be much more logical to have all the environmental variables in one place and I should take a minute to see if it can be achieved easily. IIRC Prisma did some work on this since I set this project up so maybe it is more effortless now.

prisma/scripts/preBuild.ts
```diff
@@ -2,6 +2,9 @@ import fs from "fs";
 import path from "path";
 import prisma from "../client";
 
+// Include Prisma's .env file as well, so we can fetch the DATABASE_URL
+require('dotenv').config({path: 'prisma/.env'});
+
```
Sendouc, line 7: It is strange that this is required on your end 🤔 For me as long as I have `.env` file with `DATABASE_URL` defined inside the `prisma` folder this just works.

So just to clarify the way it works for you right now is that even with the .env set like defined above this line is still required? If you exclude the line what is the error?

Sendouc:
Good clarifications. Just need a bit more info on how you have it set up currently. After merging this I will separately look into how to get it all running from one `.env` file and document that.

README.md
```diff
@@ -59,18 +65,20 @@ DATABASE_URL=postgresql://sendou@localhost:5432
...
 In addition to the steps above the steps below enable logging in.
 
-7. Create a file called `.env.local` in the root folder. In it you need following variables:
+9. Create a file called `.env.local` in the root folder. In it you need following variables:
 
 ```
-DISCORD_CLIENT_ID=
-DISCORD_CLIENT_SECRET=
-JWT_SECRET=
+DATABASE_URL="<the same DATABASE_URL you set up before>"
```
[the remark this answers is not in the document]
rofreg: Good call – I added this line when trying to resolve the issue with `npm run prebuild`, and I (wrongly) thought it was a required part of the solution. I'll update to remove it!

prisma/scripts/preBuild.ts
rofreg: Yeah, this definitely confused me as well. This line seems like it should be unnecessary, but unfortunately when I run `npm run prebuild` without it, `schema.prisma` complains that it cannot find the `DATABASE_URL`:

```
rofreg@Ryans-MacBook-Pro sendou.ink % npm run prebuild

> sendou.ink@1.0.0 prebuild
> ts-node prisma/scripts/preBuild.ts

PrismaClientInitializationError2 [PrismaClientInitializationError]:
Invalid `prisma.user.findMany()` invocation:


  error: Environment variable not found: DATABASE_URL.
  -->  schema.prisma:4
   |
 3 |   // TODO: should use same .env system as Next.JS
 4 |   url      = env("DATABASE_URL")
   |

Validation Error Count: 1
    at cb (/Users/rofreg/Dropbox/Personal/sendou.ink/node_modules/@prisma/client/runtime/index.js:34780:17)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  clientVersion: '2.19.0',
  errorCode: undefined
}
(node:77884) UnhandledPromiseRejectionWarning: Error: error: Environment variable not found: DATABASE_URL.
  -->  schema.prisma:4
   |
 3 |   // TODO: should use same .env system as Next.JS
 4 |   url      = env("DATABASE_URL")
   |

Validation Error Count: 1
    at LineStream.<anonymous> (/Users/rofreg/Dropbox/Personal/sendou.ink/node_modules/@prisma/client/runtime/index.js:29449:31)
    at LineStream.emit (events.js:314:20)
    at LineStream.EventEmitter.emit (domain.js:483:12)
    at addChunk (_stream_readable.js:298:12)
    at readableAddChunk (_stream_readable.js:273:9)
    at LineStream.Readable.push (_stream_readable.js:214:10)
    at LineStream.Transform.push (_stream_transform.js:152:32)
    at LineStream._pushBuffer (/Users/rofreg/Dropbox/Personal/sendou.ink/node_modules/@prisma/client/runtime/index.js:26388:19)
    at LineStream._transform (/Users/rofreg/Dropbox/Personal/sendou.ink/node_modules/@prisma/client/runtime/index.js:26382:10)
    at LineStream.Transform._read (_stream_transform.js:191:10)
(node:77884) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:77884) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

My dotenv files are currently:

```
# .env
DISCORD_CLIENT_ID=
DISCORD_CLIENT_SECRET=
JWT_SECRET=
NEXTAUTH_URL=http://localhost:3000

# .env.local
DISCORD_CLIENT_ID=<real value>
DISCORD_CLIENT_SECRET=<real value>
JWT_SECRET=<real value>

# prisma/.env
DATABASE_URL=postgresql://rofreg@localhost:5432/sendou_ink
```

rofreg: Oh, a few more possible clues here:
- If you add `DATABASE_URL` to `.env` (not to `prisma/.env`), then run `npm run prebuild`, it will run just fine even without my added line of code. 
- If you add `DATABASE_URL` to `.env.local`, then run `npm run prebuild`, it will fail with the same error shown above.

So it seems like `npm run prebuild` is successfully picking up the ENV vars in `.env`, but not in `.env.local` or `prisma/.env`, at least in my testing. I have no idea why, though!

Sendouc: That's a bit weird yeah. I think I will just merge this for now as it doesn't seem to be doing any harm either. I will need to see about it when I take a look at #322

Issue #85, opened 2020-09-04
Cypress tests


Sendouc:
Should at least have happy path testing for everything on the site.

PR #314, opened 2021-03-28
Ignore test run output from Cypress

rofreg:
Related to #85. Right now, running `npm run cy:run` creates output screenshots and/or videos that might accidentally get committed to Git.


Sendouc:
True! Good catch. Thanks.

Issue #197, opened 2020-12-10
Seed db for dev/testing scripts


Sendouc:
Should be able to run a command `npm run seed` and have a populated database ready for developing.

Data models can be found here: https://github.com/Sendouc/sendou.ink/blob/main/prisma/schema.prisma

PR #315, opened 2021-03-28
Use a factory to create seed data

rofreg:
Part of #197. For now, I've only updated the `User` model to use a factory. If this approach makes sense to others, then in the future I hope to finish converting the other existing fixtures into factories, and then to add more factories to complete the seed data.

As I see it, there are two main advantages to using factories instead of the existing JSON fixtures:
- Type checking: because we are generating real `User` objects, we can validate that all required attributes have been included.
- Ease-of-use: factories make it easy to provide default values for an object, and then to override those defaults when desired. (For an example, see how the "Sendou" and "NZAP" users are created in `seedUsers`.) This is also convenient when you add a new attribute to an existing data model: you can just update the factory, and all instances of that factory will automatically get the new attribute.

P.S. I have a lot of experience writing code in Ruby, Swift, etc, but not very much experience writing modern Javascript, and part of my interest in contributing to this project is to improve my skills in other languages. If there are places where my code is wrong or "un-Javascript-like", please let me know – I'd love to learn!


package.json
```diff
@@ -38,6 +38,7 @@
     "@trpc/react": "^4.0.0",
     "@trpc/server": "^4.0.0",
     "countries-list": "^2.6.1",
+    "fishery": "^1.2.0",
```
Sendouc, line 41: I believe this should be under devDependencies since we won't need it in production: (so just running `npm i fishery -D` should do the trick)

prisma/seed.ts
```diff
@@ -6,29 +6,12 @@ import {
   getPlusSuggestionsData,
   getPlusStatusesData,
 } from "./mocks/plus";
-import { getUsersData } from "./mocks/user";
+import userFactory from "../utils/factories/user"
```
Sendouc, line 9: I might prefer to keep the factories under the `prisma` folder as they are something we never need unless we are seeding data. So maybe the path could just be `./factories/user` here. What do you think? I'm not quite sure what is the best but having them close to the `seed.ts` file seems the best to me when taking the first look.

prisma/seed.ts
```diff
@@ -54,6 +37,45 @@ function throwIfNotLocalhost() {
...
+    prisma.freeAgentPost.deleteMany({}),
+    prisma.team.deleteMany({}),
+    prisma.ladderPlayerTrueSkill.deleteMany({}),
+    prisma.ladderMatchPlayer.deleteMany({}),
+    prisma.plusVotingSummary.deleteMany({}),
+    prisma.plusSuggestion.deleteMany({}),
+    prisma.plusStatus.deleteMany({}),
+    prisma.user.deleteMany({})
+  ])
+}
```
Sendouc, line 56: I figured out how we can get rid of this function. We can add a new script to the package json: `"migrate:reset": "prisma migrate reset",` (check https://www.prisma.io/docs/concepts/components/prisma-migrate)

This resets the DB and runs the seeding script.

utils/factories/user.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export default Factory.define<User>(({ sequence, onCreate }) => {
+  onCreate(user => {
+    return prisma.user.create({ data: user });
+  });
+
+  return {
+    id: sequence,
+    discordId: sequence.toString().padStart(17, '0'),
+    discordAvatar: "???",
```
Sendouc, line 13: We can also leave this null as `???` can never be the value for this. Then special case can be used for the N-ZAP user of `f809176af93132c3db5f0a5019e96339` that's their avatar value that I never change it so it can be used for testing like this.

prisma/seed.ts
rofreg: Oh nice! Yep, that definitely simplifies this code a bit 👍

rofreg:
Comments addressed! And you're totally right on the typing. I got confused because I also started to look at replacing `cypress/fixtures/user.*.json` with this new factory, but then I decided it'd be best to leave that alone for now.

In the future, factories may also be useful for automated tests – that's actually the primary way that I use factories in my other work. Factories make it easy to generate fake data for lots of different test cases, which is great when you want to test that a feature works properly for any given user configuration!

Sendouc:
Nice this is great! Yeah I'm curious what the best approach would be regarding test data. Currently I'm just using fixtures to do some semi-hacky approach for doing log in. 

Currently my approach to test data has just been testing everything against the state as defined in `seed.ts`. You are right that using factories might enable testing different states much more easily.

Issue #293, opened 2021-03-11
User page: Dividers showing up even with no bio


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/110756967-84fcf980-8253-11eb-94e9-4cb4ed91954f.png)

DoubleCookies:
This issue is pretty tricky. With empty bio there is only divider before build section (if user have them), but if bio have smth like spaces or invisible separators - code add additional divider before bio. Probably dividers work as intended.

Sendouc:
I think it's still an issue as it just looks off to the user (like maybe bio didn't load or something).

Of course the question is if we should be saving bio even when it only contains empty characters (or maybe it's not a problem?) but a quick fix could be maybe checking if `bio.trim() === ""` and then not rendering the <Divider />.

PR #333, opened 2021-03-31
Check bio length after trim()

DoubleCookies:
#293


Sendouc:
That seems to do the trick. Thanks!

Issue #330, opened 2021-03-31
Fix tests


Sendouc:
Currently most of the tests fail. The error is:

```
PrismaClientUnknownRequestError2 [PrismaClientUnknownRequestError]: 
Invalid `prisma.plusStatus.findMany()` invocation:


  Error occurred during query execution:
ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState("0A000"), message: "cached plan must not change result type", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("plancache.c"), line: Some(718), routine: Some("RevalidateCachedQuery") }) }) })
    at cb (/Users/sendou/Documents/Programming/sendou-ink/node_modules/@prisma/client/runtime/index.js:34778:17)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async ProcedureWithoutInput.call (/Users/sendou/Documents/Programming/sendou-ink/node_modules/@trpc/server/dist/router-830db415.cjs.dev.js:59:20) {
  clientVersion: '2.19.0'
}
```

With some Googling could be related to the new way of seeding the DB between tests:

```
beforeEach(() => {
  // TODO: use database transactions, instead of dropping and recreating the database with each individual test
  cy.exec("npm run migrate:reset -- --force");
});
```

Probably tests are otherwise broken too but this is the big problem to figure out for now.

rofreg:
Yup, it looks like dropping and recreating the DB between tests makes the server's Postgres connection upset. I'll look into switching back to `npm run seed` for Cypress.

PR #334, opened 2021-03-31
Start to fix tests + add more factories

rofreg:
Part of #330 – this gets more tests passing, though not yet all. There's still more work to be done to either (1) make the factory data look exactly like the old fixtures, or (2) update the tests to use more dynamic factory data instead of assuming hard-coded values from fixtures.

Apparently `npm run migrate:reset` interacts poorly with Postgres, and effectively corrupts the connection of the running dev server such that you have to restart it. As a result, running `npm run migrate:reset` between each test was causing lots of problems.


prisma/factories/plus_status.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default Factory.define<PlusStatus>(({ params, onCreate }) => {
+  onCreate(plusStatus => {
+    return prisma.plusStatus.create({ data: plusStatus });
+  });
+
+  return {
+    userId: userFactory.build().id,
+    membershipTier: 1,
+    region: _.sample([PlusRegion.EU, PlusRegion.NA]) || PlusRegion.NA,
```
rofreg, line 15: I thought `_.sample([PlusRegion.EU, PlusRegion.NA])` would be enough here, but I guess it can technically return `undefined` as well? There might be a better way to make TypeScript happy here.

That said, this is a good example of how to make a factory more dynamic and random (which can be useful for catching edge cases in your automated tests!)

cypress/support/index.ts
```diff
@@ -14,5 +14,5 @@ Cypress.Commands.add("login", (user: "sendou" | "nzap") => {
 
 beforeEach(() => {
   // TODO: use database transactions, instead of dropping and recreating the database with each individual test
-  cy.exec("npm run migrate:reset -- --force");
+  exec("npm run seed");
```
Sendouc, line 17:
```suggestion
  cy.exec("npm run seed");
```

prisma/factories/plus_status.ts
```diff
@@ -0,0 +1,22 @@
+import { Factory } from "fishery";
+import { PlusStatus, PlusRegion } from "@prisma/client";
+import prisma from "../client";
+import userFactory from './user';
+import _ from "lodash";
```
Sendouc, line 5: I was actually kind of surprised lodash is available since it's not a dependency. I guess we can still use dependencies of dependencies like this? Probably is fine but maybe adding it explicitly as devDependency is the best.

prisma/factories/plus_status.ts
Sendouc: I think the method might be mistyped actually. Reading through the [docs](https://lodash.com/docs/#sample) there should be no way it returns `undefined` or `null` when given an array even if the types say so.

So in this case I think in this case the most TypeScript friendly way would actually be coercing the value: 
```suggestion
    region: _.sample([PlusRegion.EU, PlusRegion.NA])!,
```

Using the `||` operator kind of means to me it can sometime be falsy and we are preparing for it while actually we just have more knowledge than TypeScript about this case so coercing it is more clear.

As a side note if we were preparing for a falsy value here the modern JS way of doing it would be via the [?? operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). It's better than using `||` because it only gives the right-hand value if `undefined` or `null` and not for example with `0` that's also a falsy value in JS.

Sendouc:
For file naming camelCasing so use prisma/factories/plus_status.ts -> prisma/factories/plusStatus.ts

Once we get this merged I can fix the existing tests. The Plus Server side of the site has a bunch of edge cases anyway so I think it might be hard to figure out what's going on for someone else.

Still a fan of this though! It's nice once we get some more dynamic seeded data.

prisma/factories/plus_voting_summary.ts
```diff
@@ -0,0 +1,21 @@
...
+import prisma from "../client";
+import userFactory from './user';
+
+export default Factory.define<PlusVotingSummary>(({ params, onCreate }) => {
+  onCreate(plusVotingSummary => {
+    return prisma.plusVotingSummary.create({ data: plusVotingSummary });
+  });
+
+  return {
+    userId: userFactory.build().id,
```
Sendouc, line 12: Just curious but what is happening with this line? 🤔

prisma/seed.ts
```diff
@@ -36,25 +37,40 @@ function throwIfNotLocalhost() {
...
 }
 
-async function seedNewData() {
-  await seedUsers();
-  await prisma.plusStatus.createMany({ data: getPlusStatusesData() });
-  await prisma.plusSuggestion.createMany({ data: getPlusSuggestionsData() });
-  await prisma.plusVotingSummary.createMany({
-    data: getPlusVotingSummaryData(),
-  });
+async function dropAllData() {
```
Sendouc, line 40: The additional challenge here is that everything has to be deleted in the right order otherwise Postgres complains. I think the easiest way for now is just to change the order of these as needed.

Also maybe worth it to see the discussion here for alternative solutions: https://github.com/prisma/docs/issues/451

Sendouc (changes requested):
Overall looks good! 🦖

prisma/factories/plus_status.ts
rofreg: Ah, yeah, this makes total sense – I didn't even check if was already included. I can add it as a devDependency, unless you think there's a better way to grab a random item from an array.

prisma/factories/plus_voting_summary.ts
rofreg: This definitely needs further work 🙂 The idea here is that factories can automatically create other required associations, so you could theoretically do something like:
```typescript
// We can manually specify the User for a PlusVotingSummary if we want to...
const user = await userFactory.create();
const plusVotingSummary = await plusVotingSummaryFactory.create({user: user});
console.log(plusVotingSummary.userId); // user.id

// ...but by default, the PlusVotingSummary factory will _automatically_ create an associated User for us
const plusVotingSummary = await plusVotingSummaryFactory.create();
console.log(plusVotingSummary.userId); // a new User id that has automatically been saved to the DB
```

That said, I'm definitely not doing this correctly yet. I think I'll revert this bit of code and used a hard-coded value for now.

prisma/factories/plus_status.ts
Sendouc: Yeah I think I will write a custom helper for this. See if we can also make Typescript happy without type coercion that way. 

Generally not against using Lodash just have gotten the impression most of its claim to fame was from a time when JS was more limited and personally I haven't felt the need to use it.

PR #341, opened 2021-04-01
Try out GitHub Actions

rofreg:
I'm opening this PR as a test – I don't expect it to work first try. (Opening the PR should cause GitHub Actions to run.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,12 @@
+name: CI
+on:
+  - pull_request
+jobs:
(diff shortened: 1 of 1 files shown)


Sendouc:
Lemme just merge this let's see what happens.

For now I think we need couple things to happen with this:

1) Run Cypress tests
2) https://prettier.io/docs/en/cli.html#--check

Later ESLint (this is coming to Next.JS really soon so I'd rather wait their official implementation here). Are there others?

Issue #321, opened 2021-03-30
Pipeline


Sendouc:
Once a commit is pushed or PR opened should ran through a pipeline:

1) Check against ESLint #319
2) Check against Prettier
3) Run Cypress tests

rofreg:
I've used GitHub Actions for this on other projects, and it's been very convenient – and free for public repos, which is great. Cypress even appears to have [its own custom GitHub Action](https://github.com/cypress-io/github-action#basic). Maybe I'll look into this soon!

Sendouc:
Yup sounds great! I'm a fan of GitHub Actions from what I know of it. It's nice to have everything related to the development of this project in one place as much as possible.

If you have any questions about the project you can either just make an issue with the tag question or ask on the [Discord server](https://discord.gg/sendou). I check both actively so should be able to answer quick.

PR #342, opened 2021-04-01
Further testing of GitHub Actions

rofreg:
Part of #321. Follow-up to PR #341.

In theory, opening this PR should cause [the `ci.yml` action](https://github.com/Sendouc/sendou.ink/actions/workflows/ci.yml) to run. Then we can see whether the GitHub Action actually works! Again, I suspect this will actually fail on first try 😛 

Once this is working and all the tests are passing, the [repo settings](https://github.com/rofreg/sendou.ink/settings) can be updated such that any merge into the `main` branch requires CI to pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,10 @@ jobs:
         uses: cypress-io/github-action@9eab5368cd2520a946489cd3f937583ff5a30443 # v2
         with:
           start: npm run dev
+      - name: Save Cypress artifacts
+        uses: actions/upload-artifact@ea3d524381d563437a7d64af63f3d75ca55521c4 # v2
+        with:
+          name: cypress-outputs
+          path: |
+            cypress/screenshots
+            cypress/videos
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,11 +5,27 @@ jobs:
   cypress-run:
     name: Cypress tests
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_PASSWORD: password
+          POSTGRES_USER: sendou_ink
+        ports:
+          - 5432:5432
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


rofreg:
🎉 Looks like this sorta works! `plus.spec.ts` is broken, but `calendar.spec.ts` passes, so I think GitHub Actions is properly set up to boot the server and run Cypress 🚀

Sendouc:
Oh this is really nice! 🚀 

Would it be possible to have it set up in a way where we can download fragments (screenshots / videos) of Cypress runs? The Plus Server functionality related tests work for me locally so I'd want to try to understand what is happening there.

As far as I can tell the only test that passes is just testing the lack of something so it might very well be a completely blank page passes even with the DB connection failing.

rofreg:
It looks like it's possible for GitHub Actions to save artifacts after each run: https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts. I haven't done that before, but I'll give it try later today/this weekend.

Also your diagnosis sounds right to me, lol – maybe GitHub Actions isn't actually successfully starting the server yet. If so, screenshots / videos should be helpful to diagnose the problem!

rofreg:
Cypress output is now being saved! (Click "Details" next to "CI / Cypress tests" below, and then click "Artifacts" in the upper-right of the page.)

It looks like `messages.js` is failing to load, though I'm not immediately sure why that's failing on GitHub Actions and working locally. I'll try a few things, but I might need help diagnosing the issue.

rofreg:
Most tests are passing now! 🎉 I think the remaining failures may be legitimate test failures – if so, it might make sense to merge this PR as-is, and fix those issues in a separate PR.

Sendouc:
Nice! 🎉 

Yeah made a separate issue for fixing the tests. Loading the files from JSON is not what the i18n we use recommends but as long as it works for now it's good enough. I need to switch out the lib to another one anyway (will detail this in another issue).

Sendouc:
Had to revert this (check https://github.com/Sendouc/sendou.ink/pull/361 for details)

Not sure if I can actually reopen this PR or does there have to be a new PR?

Not 100% sure how we want to fix it. I suppose we could use `npm run compile && npm run dev` to start the app for tests? I will be replacing the i18n library asap since it is causing some severe problems but just to get this merged since having a pipeline like this is some good value.

rofreg:
Ah, so sorry for the trouble! Reverting for now definitely makes sense – if there's a simple workaround to get the existing i18n working w/ GitHub Actions, that's probably a good place to start, and then we can always replace the i18n at a later time if desired.

Issue #297, opened 2021-03-16
Maplist Generator: custom order


Sendouc:
"would be nice if the map generator had a couple other features
generate maps in certain order (aka i want rm sz tc cb for example)"

In addition to "All modes equally" and "SZ every other" add an option for a customized order. Should be able to omit modes. Might be nice to be able to have modes more than once (e.g. I want a list of maps that goes like SZ->TC->SZ->RM->SZ->TC...)

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/346

PR #346, opened 2021-04-02
Maplist generator: Custom Order

DoubleCookies:
#297 
image
I think it is finished.
### Updates
- Created MultipleModeSelector. It looks similar to MySelect but has different handling options logic for allowing duplicates (which is a bit weird) and modes icons _(yay)_. I thought about putting everything in one file (MySelect) but with it component will be too complicated.
- Added new radio mode "Custom Order".
- Shuffling modes array is after filtering (and only in non-custom order)
- changing modes array is after getting value (without it generation starts only from second mode in array).


pages/maps.tsx
```diff
@@ -347,9 +360,23 @@ const MapsGeneratorPage = () => {
...
+        options={[
+          { label: "Splat Zones", value: "SZ", data: "SZ" },
+          { label: "Tower Control", value: "TC", data: "TC" },
+          { label: "Rainmaker", value: "RM", data: "RM" },
+          { label: "Clam Blitz", value: "CB", data: "CB" },
+        ]}
+        isDisabled={generationMode !== "CUSTOM_ORDER"}
+        setValue={getModeValues}
+        width={"90%"}
+      />
```
Sendouc, line 378: This should only be rendered if Custom order is selected I think.

pages/maps.tsx
```diff
@@ -127,15 +131,22 @@ const MapsGeneratorPage = () => {
...
+      (mode) => modeStages[mode].length > 0
+    );
+    const modes =
+      generationMode === "CUSTOM_ORDER"
+        ? filteredModes
+        : shuffleArray(filteredModes);
     if (modes.length === 0) {
-      return "I can't generate a maplist without any maps in it you know.";
+      return generationMode === "CUSTOM_ORDER"
+        ? "I can't generate a maplist without any mode you know."
```
Sendouc, line 147:
```suggestion
        ? "I can't generate a maplist without any modes you know."
```

Sendouc (changes requested):
Just a few minor things but otherwise really solid work. Good job!

pages/maps.tsx
DoubleCookies: Updated

Issue #298, opened 2021-03-16
Light mode borders too similar color to background


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/111335657-3b5e4580-867d-11eb-97d0-f69ddae32f7c.png)

Example of how Chakra UI defaults can be overwritten: https://github.com/Sendouc/sendou.ink/blob/main/pages/_app.tsx#L56

PR #355, opened 2021-04-03
borderColor fix

### LINKED ISSUE #347: MySelect component has wrong border color in light mode
image

Should be same as the Chakra-UI's Select border color. Seems to be correct in dark mode.

DoubleCookies:
#347 
It solves problem now but probably in the long run there should be another solution (#298 mentions about it).


Sendouc:
Yeah that's a good improvement for now. Thanks 👍

Issue #338, opened 2021-04-01
Build analyzer: Jr ISS bomb consumption wrong


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/113335369-83dc5b00-932d-11eb-9250-a01e827da80d.png)

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/356#issuecomment-813435749

PR #356, opened 2021-04-03
fix calcs for JR

DoubleCookies:
#338 
_Math is weird._ Ink consumption for sub weapons now is the same compared to loadout.


Sendouc:
I think loadout.ink (or the updated fork of it) has to be wrong here. Look at the video posted here: https://twitter.com/Sey_spl/status/1377463957850480641

According to analyzer currently with that set Splat Bombs take 50.38% of ink tank and according to loadout.ink fork it takes 49.873%. However we can easily go into the test range while having 1.4 ISS and try to double tap R -> it gives the message of no ink. So that's why I think simple the analyzer is correct currently.

Either way this is a good cross reference to have in case we get similar issues with the analyzer again so thank you 👍

PR #357, opened 2021-04-03
suggested changes for trpc 5

KATT:
note that this is an alpha release i've upgraded to

what do you think about this? feel free to make suggestions over at https://github.com/trpc/trpc/pull/244


Sendouc:
That makes sense to me yeah. I saw you on Twitter as well wondering if useRef() and using useState with function are equivalent in this case but unfortunately I don't know.

Sendouc:
Everything seems to be working. Thanks for the release and PR 👍

Issue #312, opened 2021-03-27
Firefox / mobile weird block in weapon selector when nothing selected


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/112737434-43629300-8f63-11eb-9e66-da918c559121.png)

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/363

PR #363, opened 2021-04-05
Fix for WeaponSelector in Firefox

DoubleCookies:
#312 
I found that it is default behavior for Firefox if image doesn't exists (in this case - it return 404 if no weapon is selected). We can just don't render image if nothing is selected.
(Another option - use placeholder image (with is fully transparent) but first solution works fine)


Sendouc:
Nice fix! This has been pretty annoying so nice to have it fixed. Just changed your solution a bit to return early since I personally think that's more React-y.

PR #369, opened 2021-04-07
Re-enable GitHub Actions

rofreg:
Reverts #361 / re-enables most of #342.

I'm opening this PR as a starting place – we'll see how it goes! It will probably require more tweaking and testing before the GitHub Action fully passes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "npm" 
+  - package-ecosystem: "npm"
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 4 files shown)


Sendouc:
Isn't this what we want?

- Tests pass except Plus Server tests (which need more investigating)
- We are using the compiled i18n files

As far as I can tell this is it but can you confirm? 🕵️

rofreg:
@Sendouc Lol, you're totally right – that's what I get for working on this right before bed 😛 Should be good to merge, and then we can fix the Plus Server tests in a later PR!

Sendouc:
Seems like a plan!

Issue #388, opened 2021-04-08
Popover size too wide


Sendouc:
Chakra-UI 1.5 broke them probably:

![image](https://user-images.githubusercontent.com/38327916/114056114-028b4800-989a-11eb-8613-65b72bf9f89f.png)

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/393

PR #393, opened 2021-04-09
Unset width for popover in navigation

DoubleCookies:
#388 
It allows to ignore chakra-ui variable for width (20rem) and use width depending on content size.
Also `min-width: max-content` can be useful if smth goes wrong with this solution.


DoubleCookies:
I just understood that every `PopoverContent` should be updated as well.

Sendouc:
Maybe it can be given as a default prop? Like your Heading fix.

DoubleCookies:
Updated. Now it's finished.

Sendouc:
Nice fix. Good one.

Issue #389, opened 2021-04-08
React-Select better searching


Sendouc:
Currently searching for a weapon is pretty rigid:

![image](https://user-images.githubusercontent.com/38327916/114057266-15524c80-989b-11eb-855b-911f786e15b1.png)

See: https://github.com/JedWatson/react-select/issues/3067#issue-363771398

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/397

PR #397, opened 2021-04-09
Custom filter option

DoubleCookies:
#389 
Suggested solution works fine - can be tested in weapon and gear selectors.


Sendouc:
Perfect! Thanks.

Issue #233, opened 2020-12-21
User Selector: Can't search if JP characters + search via twitter too?


Sendouc:

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/413

PR #413, opened 2021-04-13
UsersSelector - search via twitter

DoubleCookies:
#233 
I didn't find problems with JP characters but implemented second part of issue. Also made customFilterOption as props for MySelect (because different selectors require different logic).


Sendouc:
Nice 👍  Looks good.

Issue #404, opened 2021-04-11
Modals: Broken spacing


Sendouc:
Seems to be broken after update to Chakra UI:

![image](https://user-images.githubusercontent.com/38327916/114301303-eb806c00-9acc-11eb-8e87-d2b9b635808c.png)

(the Form Labels should not be connected to the previous input like that but instead there should be spacing between them)

Was previously handled by setting marginTop on FormLabel:

```jsx
<FormLabel htmlFor="date" mt={4}>
  <Trans>Date</Trans>
</FormLabel>
```

But this seems to be no longer supported.

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/417

PR #417, opened 2021-04-16
Modals Spacing

DoubleCookies:
#404 
I think form labels + controllers should be wrapped in formcontrol (with it old form labels still have margin). Probably there is another solution but this looks logical (another elements like markdowntextarea have formcontrol and form label inside - margins are working).


Sendouc:
Seems good. Thank you!

Issue #394, opened 2021-04-09
Switch themed color


Sendouc:
Should work the same as radio button (so themed globally):

![image](https://user-images.githubusercontent.com/38327916/114138483-5f2c4880-9916-11eb-90ea-4855d8dd4d6f.png)

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/418

PR #418, opened 2021-04-16
Themed switch

DoubleCookies:
#394 
"It just works"


Sendouc:
It's my favorite when things just work.

PR #428, opened 2021-04-21
Add ID and profile updatedAt to api/bot/freeagents

Alecat:


prisma/migrations/20210421084805_add_profile_updated_at/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "Profile" ADD COLUMN "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT NOW();
```
Alecat, line 2: Should it instead use epoch time / something else for existing. entries?

Alecat:
Closing - will handle ID and updatedAt schema change as separate commits.

PR #444, opened 2021-04-26
⬆️ upgrade to trpc6

KATT:
doing this as part of my due diligence for v6 version


KATT:
will release v6 later this week - here's the branch https://github.com/trpc/trpc/pull/306

it has automatic SSR-support - not sure how it works with next-auth yet though and the cold starts can be painful if you have a lot of unique pages

Sendouc:
The stuff in the PR looks clean. For sure will make it easier for people to get involved with your library. Not too sure if I wanna migrate to using `getInitialProps()` for everything though. I need to look into it more but based on my current knowledge I don't think it's quite what I want.

KATT:
Ah yeah, missed answering to this, it only wraps getInitialProps if you enable ssr

Issue #433, opened 2021-04-22
Mobile menu: Clicking My Page should close the menu


Sendouc:
https://user-images.githubusercontent.com/38327916/115707730-d252a780-a377-11eb-94dc-b6cf6311d0b9.mp4

Currently it doesn't close the menu so it looks like it doesn't do anything. Other buttons work correctly.

PR #447, opened 2021-04-27
close menu after click on user page

DoubleCookies:
#433


Sendouc:
Thank you but it was already done!

Issue #439, opened 2021-04-24
Maplist generator: link should retain settings


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/115952015-7e1d0400-a4ec-11eb-889e-66864124d389.png)

Now only retains the pool.

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/448

PR #448, opened 2021-04-28
Saving maplist generator settings in URL

DoubleCookies:
#439 
It saves count, generation mode and custom order modes.


Sendouc:
Nice job!

Issue #387, opened 2021-04-08
validateDOMNesting warning when using custom emojis in Markdown


Sendouc:
react_devtools_backend.js:2430 Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.

To reproduce:

This happens whenever you use a "custom emoji" in your markdown. For example just have user profile of: 

`:bloblobber:`

When the WeaponImage renders this warning is displayed.

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/496

PR #496, opened 2021-05-12
Fix validateDOMNesting warning

DoubleCookies:
#387 
Rendering div instead of p for markdown should be enough.


Sendouc:
Yeah seems to work. Thank you.

Issue #429, opened 2021-04-21
Should be able to scroll side navigation indepently


Sendouc:
Currently if you are on a long page but your screen isn't large enough to fit all the side nav icons on screen so you want to scroll down this can only be achieved by scrolling down on the main page.

Instead you should be able to scroll down the side navigation independently from the main content.

Probably a matter of inserting the correct CSS rule or two.

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/497

PR #497, opened 2021-05-13
Scrollable nav

DoubleCookies:
#429 
I disable scrollbar visibilty but scroll is working. Max-height probably can be changed.


Sendouc:
Nice one. Good job.

Issue #488, opened 2021-05-09
Weapon selector allow searching by sub or special weapon


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/117580247-cb5dc000-b0ff-11eb-9903-06dcab42b44d.png)

If exact match for example for "Inkjet" (case insensitive) show all weapons with it. Maybe for English initially then later for all languages supported?

Red-CS:
Has anybody started this? And where would this feature be implemented? Along the sidebar with Browser, Tier list, etc., or built into something else?

Sendouc:
If it's not assigned to anyone and nobody commented anything you can think it's up for grabs. The change needs to be made to the `WeaponSelector` component.

DoubleCookies:
I can try to implement this if you wouldn't mind.

Sendouc:
I'd say go unless @Red-CS started already

Red-CS:
> I can try to implement this if you wouldn't mind.

I haven't started, so go ahead, all yours!

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/503

PR #503, opened 2021-05-18
Search weapons by sub or special (en-only)

DoubleCookies:
#488


Sendouc:
This is neat. Good work.

Issue #518, opened 2021-05-21
Builds: ability / mode dropdowns show icons


Sendouc:
For the filtering dropdowns have them use react-select and render some icon.

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/523

PR #523, opened 2021-05-27
New filters for builds page

DoubleCookies:
#518

### DIFF
--- a/components/builds/BuildFilters.tsx
+++ b/components/builds/BuildFilters.tsx
@@ -9,14 +9,13 @@ import {
   NumberInputField,
   NumberInputStepper,
   Radio,
-  Select,
 } from "@chakra-ui/react";
 import { t, Trans } from "@lingui/macro";
 import { Ability, Mode } from "@prisma/client";
 import AbilityIcon from "components/common/AbilityIcon";
 import { UseBuildsByWeaponDispatch, UseBuildsByWeaponState, } from "hooks/builds";
 import { useMyTheme } from "hooks/common";
-import { Fragment, useState } from "react";
+import React, { Fragment, useState } from "react";
 import { FiTrash } from "react-icons/fi";
 import { abilities, isMainAbility } from "utils/lists/abilities";
 import { components } from "react-select";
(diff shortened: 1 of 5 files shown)


components/builds/BuildFilters.tsx
```diff
@@ -19,16 +18,98 @@ import {
...
+      <Flex alignItems="center">{props.data.label}</Flex>
+    </components.SingleValue>
+  );
+};
+
 const BuildFilters: React.FC<Props> = ({ filters, dispatch }) => {
+  const [mode, setMode] = useState<{ label: string; value: string }>(
+    modeOptions[0]
+  );
   const [modeValueChanged, setModeValueChanged] = useState(false);
```
Sendouc, line 113: This state is never used. I guess it can just be removed as it's not too important to filter out "All modes" from the start ( the reason why it was originally there)

components/builds/BuildFilters.tsx
```diff
@@ -19,16 +18,98 @@ import {
...
+    </components.Option>
+  );
+};
+
+const AbilityOption = (props: any) => {
+  return (
+    <components.Option {...props}>
+      <Flex alignItems="center">
+        <Box mr="0.5em">
+          <AbilityIcon ability={props.value} size={"SUBTINY"} />
```
Sendouc, line 86: Useless brackets (will setup a linter soon that will catch these type of issues but bear with me for now 💦 )

Sendouc:
See a few comments. Otherwise a few stylistic nitpicks:

1) All modes is more indented to the right than Filter by ability here:
image

Should have same margin.

2) Would it be possible to make the dropdowns smaller? I don't think we have tried that with react-select before but I imagine it's trivial. It'd be perfect if they would be about the same size as they were previously when done with Chakra UI dropdowns.

Good job otherwise!

DoubleCookies:
Fixed everything except the last point (probably will finish it tomorrow)

DoubleCookies:
Guess it is done. Selects on builds page are a bit smaller now and can be styled.

Sendouc:
I really like how this turned out. Great work! 🎉

PR #535, opened 2021-06-08
Update-heb-messages

shachar700:
Small changes to translations and added translation to new lines


Sendouc:
Cool. Thank you. Btw translations are in bit of a transition period. Waiting for some libraries to update then I will be adding translations to more pages (some of the newer pages lack them currently).

Issue #532, opened 2021-06-03
Builds Selector: search by brand


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/120690776-4f605900-c4ae-11eb-892f-b8b3dc193ddc.png)

So like how it works currently with weapon selector and sub/special weapons.

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/541

PR #541, opened 2021-06-11
Search by brand

DoubleCookies:
#532


Sendouc:
Good work as always. Thank you!

Issue #548, opened 2021-06-16
Markdown: Code should have scrollbars


Sendouc:
Currently just takes the full width of space. Instead the max width should be restricted and scrollbar shown if content is too large.

![image](https://user-images.githubusercontent.com/38327916/122246741-7ddb2c80-cecf-11eb-9859-c48708398212.png)

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/559

PR #559, opened 2021-06-18
Scroll for code block

DoubleCookies:
#548


Sendouc:
That works. Thanks. Seems like the page is still zoomed out on mobile but that's probably unrelated.

Issue #570, opened 2021-06-26
Team Splitter crash


Sendouc:
Shahar — Today at 02:37
In team-splitter, if you press reset and then start without adding players, it will give you this message with white screen: Application error: a client-side exception has occurred (developer guidance).

Sendouc:
https://github.com/Sendouc/sendou.ink/pull/579

PR #579, opened 2021-07-07
Block button by default

DoubleCookies:
#570 
Changed default state


Sendouc:
That seems to do the trick! Thank you.

Issue #561, opened 2021-06-19
Badges show multiplier 


Sendouc:
Multiplier instead of duplication

Sendouc:
https://github.com/Sendouc/sendou.ink/issues/561

PR #588, opened 2021-07-14
Badges multiplier near icon

DoubleCookies:
#561


Sendouc:
This is really good work. Just how I imagined it. Thanks as always!

Issue #591, opened 2021-07-18
Weapon selector allow looking up weapons via aliases


Sendouc:
"hey Sendou, it'd be cool to have a smarter suggestion filter for weapon builds let's face it does anyone write N-ZAP with a hyphen
I've written a big lookup for Dola if you'd find that useful? -- https://github.com/kjhf/DolaBot/blob/main/src/DolaBot/helpers/weapons.py
"

Would be good UX to allow people to write weapons as they are commonly known.

Copy paste if the link above 404's:

```json
WEAPONS = {
    ".52 Gal": ["gal", "52", "v52", "52g"],  # default gal
    ".52 Gal Deco": ["galdeco", "52deco", "52galdeco", "52gd"],
    ".96 Gal": ["96", "v96", "96g"],
    ".96 Gal Deco": ["96deco", "96galdeco", "96gd"],
    "Aerospray MG": ["mg", "aeromg", "silveraero", "silveraerospray", "aero", "aerospray"],  # default aero
    "Aerospray PG": ["pg", "aeropg", "bronzeaero", "bronzeaerospray"],
    "Aerospray RG": ["rg", "aerorg", "goldaero", "goldaerospray"],
    "Ballpoint Splatling": ["ballpoint", "bp", "pen"],  # default ballpoint
    "Ballpoint Splatling Nouveau": ["ballpointnouveau", "bpn", "bpsn", "bsn"],
    "Bamboozler 14 Mk I": ["bambooi", "bamboo1", "bamboo14mki", "bamboomki", "bamboomk1"],
    "Bamboozler 14 Mk II": ["bambooii", "bamboo2", "bamboo14mkii", "bamboomkii", "bamboomk2"],
    "Bamboozler 14 Mk III": ["bambooiii", "bamboo3", "bamboo14mkiii", "bamboomkiii", "bamboomk3"],
    "Blaster": ["vblaster"],
    "Bloblobber": ["blob", "vblob"],
    "Bloblobber Deco": ["blobdeco"],
    "Carbon Roller": ["carbon", "vcarbon"],
    "Carbon Roller Deco": ["carbondeco", "crd"],
    "Cherry H-3 Nozzlenose": ["cherry", "ch3", "ch3n", "cherrynozzle"],
    "Clash Blaster": ["clash", "vclash", "clashter"],
    "Clash Blaster Neo": ["clashneo", "clashterneo", "cbn"],
[...]

PR #594, opened 2021-07-20
Weapon aliases

DoubleCookies:
#591 
I think filtering aliases by substring is better than by complete match. Sometimes it will return more weapons than expected (rare example - "dd" alias) but still better than writing full name of alias.


Sendouc:
Cool! Nice work.

kjhf:
Hey that's really cool, nice one!

PR #632, opened 2021-08-08
Maplist generator: Add option to display map pool as table

inkfarer:
This PR adds an alternate method of displaying the map pool.

### Why?
I was tasked with creating a [map pool graphic](https://cdn.discordapp.com/attachments/722960400831873105/862796200897150976/lowinkmaps-jul21.png) based off a screenshot from the s.ink map list generator, which I found mildly annoying due to the way the map pool is displayed.

### DIFF
--- a/components/common/NewTable.tsx
+++ b/components/common/NewTable.tsx
@@ -14,7 +14,7 @@ import React, { Fragment } from "react";
 import { getRankingString } from "utils/strings";
 import OutlinedBox from "./OutlinedBox";
 
-type TableRow = Record<string, React.ReactNode> & { id: number };
+export type TableRow = Record<string, React.ReactNode> & { id: number };
 
 export default function NewTable({
   caption,
--- a/components/common/NewTable.tsx
+++ b/components/common/NewTable.tsx
@@ -14,6 +14,8 @@ import React, { Fragment } from "react";
 import { getRankingString } from "utils/strings";
 import OutlinedBox from "./OutlinedBox";
 
+type TableRow = Record<string, React.ReactNode> & { id: number };
+
 export default function NewTable({
   caption,
   headers,
@@ -27,7 +29,7 @@ export default function NewTable({
     name: string;
     dataKey: string;
   }[];
-  data: ((Record<string, React.ReactNode> & { id: number }) | null)[];
+  data: (TableRow | null)[];
   smallAtPx?: string;
   leaderboardKey?: string;
   size?: "sm" | "md" | "lg"
@@ -36,17 +38,15 @@ export default function NewTable({
     smallAtPx ? `(max-width: ${smallAtPx}px)` : "(max-width: 600px)"
   );
 
+  const nonNullData = data.filter((row): row is TableRow => row !== null);
+
   let lastValue: any = null;
   let rankToUse: number = 1;
 
   if (isSmall) {
     return (
       <>
-        {data.map((row, i) => {
-          if (row === null) {
-            return;
-          }
-
+        {nonNullData.map((row, i) => {
           if (leaderboardKey) {
             if (row[leaderboardKey] !== lastValue) rankToUse = i + 1;
             lastValue = row[leaderboardKey];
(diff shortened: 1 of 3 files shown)


pages/maps.tsx
```diff
@@ -53,6 +58,8 @@ const MapsGeneratorPage = () => {
   const [count, setCount] = useState(getInitialCount());
   const [editing, setEditing] = useState(false);
   const [copied, setCopied] = useState<null | "URL" | "LIST">(null);
+  const [displayType, setDisplayType] = useState<"TABLE" | "LIST">("LIST");
```
Sendouc, line 61: I think we can just do off with the select and use Table when not mobile and the Original for mobile. With current solution there is this when resizing window:

image

+ I think the new one is more illustrative anyway when the screen size allows it so allowing switching is not required.

components/common/NewTable.tsx
```diff
@@ -41,6 +43,10 @@ export default function NewTable({
     return (
       <>
         {data.map((row, i) => {
+          if (row === null) {
+            return;
+          }
```
Sendouc, line 48: It would be cleaner to first filter data for null then this check is unnecessary. If it makes TS angry I'd prefer returning `null` here. Idk what React does with `undefined` as part of an array but it's more standard to use `null` I think.

components/common/NewTable.tsx
```diff
@@ -102,7 +108,7 @@ export default function NewTable({
 
   return (
     <OutlinedBox>
-      <Table variant="simple" fontSize="sm">
+      <Table variant="simple" fontSize="sm" size={size ? size : 'md'}>
```
Sendouc, line 111:
```suggestion
      <Table variant="simple" fontSize="sm" size={size ?? 'md'}>
```

Sendouc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator meant for this use case.

components/common/NewTable.tsx
```diff
@@ -114,6 +120,10 @@ export default function NewTable({
         </Thead>
         <Tbody>
           {data.map((row, i) => {
+            if (row === null) {
+              return;
+            }
+
```
Sendouc, line 126: Same as above.

pages/maps.tsx
```diff
@@ -398,57 +405,92 @@ const MapsGeneratorPage = () => {
...
-          <Box textAlign="center">
-            {Object.entries(stagesSelected).map(([stage, modes]) =>
-              modes.includes("CB") ? (
-                <SubText key={stage}>{stage}</SubText>
-              ) : null
-            )}
-          </Box>
-        </Grid>
+      ) : (!isMobile &&
+        <>
```
Sendouc, line 409: Could consider extracting the component here. The kind of nesting we have here is starting to make the component hard to read.

pages/maps.tsx
```diff
@@ -29,9 +29,14 @@ import SubText from "components/common/SubText";
...
 import { ChangeEvent, Fragment, useEffect, useState } from "react";
 import { FiCheck, FiFilter, FiRotateCw } from "react-icons/fi";
+import { BsTable } from 'react-icons/bs';
+import { FaListUl, FaCheck } from 'react-icons/fa';
 import { shuffleArray } from "utils/arrays";
 import { stages } from "utils/lists/stages";
 import MyHead from "../components/common/MyHead";
+import { Flex } from '@chakra-ui/layout';
+import NewTable from "../components/common/NewTable";
+import {CSSVariables} from "../utils/CSSVariables";
```
Sendouc, line 39: Should be caught by the pipeline but make sure to run Prettier. I think it's a command.

pages/maps.tsx
```diff
@@ -398,57 +405,92 @@ const MapsGeneratorPage = () => {
...
+                { name: t`Clam Blitz`, dataKey: RankedMode.CB }
+              ]}
+              data={Object.entries(stagesSelected).map(([stage, modes], index) => (modes.length > 0 ? {
+                id: index,
+                stage: t({id: stage}),
+                ...(Object.values(RankedMode).reduce((map, mode) => {
+                  map[mode] = (modes.includes(mode) ? <FaCheck color={CSSVariables.themeColor} style={{ margin: "auto" }} /> : null);
+                  return map;
+                }, ({} as {[mode in RankedMode]: JSX.Element | null}))),
+              } : null))}
```
Sendouc, line 488: This is a bit hard to read. Might be worth to extract a function or two.

Sendouc (changes requested):
Hey first of all this is cool. Definitely an improved UI. Thanks a lot for the PR. Please check out the comments and let me know if you have any questions or disagree with something.

Sendouc:
Looks solid! Good job

Issue #574, opened 2021-06-30
(suggestion) Allow users to make builds with missing ability slots


jaherron:
I have some gear that I want to make builds for but they don't have all of their ability slots.
Here's a [screenshot](https://drive.google.com/file/d/1wpD4P3sJXWqYpUMPly5yb3nAyMpOKxIJ/view?usp=drivesdk) (goes to Google Drive) of the build I was trying to make.
There needs to be a way for someone to exclude abilities from builds if their gear doesn't have all of its slots.

Sendouc:
Yeah it probably makes sense to allow that. It has been suggested before as well. Although I'd probably make it so that those builds are hidden on the main /builds page still I imagine mostly people are looking for full builds. Still shown on your own page though.

Issue #666, opened 2021-09-23
Global CSS cannot be imported from files other than your Custom <App>. Failed to build.


kjhf:
./pages/styles.css
Global CSS cannot be imported from files other than your Custom <App>. Due to the Global nature of stylesheets, and to avoid conflicts, Please move all first-party global CSS imports to pages/_app.js. Or convert the import to Component-Level CSS (CSS Modules).
Read more: https://nextjs.org/docs/messages/css-global
Location: pages\_app.tsx

This error occurred during the build process and can only be dismissed by fixing the error.

-----
Full printout from the terminal:
PS D:\User\Projects\sendou.ink> npm run dev

> sendou.ink@1.0.0 dev
> next dev

ready - started server on 0.0.0.0:3000, url: http://localhost:3000
info  - Loaded env from D:\User\Projects\sendou.ink\.env
info  - Using webpack 5. Reason: Enabled by default https://nextjs.org/docs/messages/webpack5
warn  - You have enabled experimental feature(s).
warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use them at your own risk.

(node:9380) [DEP_WEBPACK_MODULE_ISSUER] DeprecationWarning: Module.issuer: Use new ModuleGraph API
(Use `node --trace-deprecation ...` to show where the warning was created)
error - ./pages/styles.css
Global CSS cannot be imported from files other than your Custom <App>. Due to the Global nature of stylesheets, and to avoid conflicts, Please move all first-party global CSS imports to pages/_app.js. Or convert the import to Component-Level CSS (CSS Modules).
Read more: https://nextjs.org/docs/messages/css-global
Location: pages\_app.tsx

-----

Node version:
node -v
v16.9.1

NPM version:
npm -v
7.24

Running on Windows 10 64-bit

Followed README instructions:
1. Clone the project (running from main)
2. Run npm i to install dependencies
3. Run npm run compile to compile translation files.
4. Run npm run dev to start the development server at http://localhost:3000/. (To stop the server at any time, type Ctrl+C.)

The server spins up however only displays the error message above.

Sendouc:
Hmm yeah like we talked on Discord this is kind of puzzling me. I could try my Windows machine to see if that has any impact. It still looks like to me that somehow TypeScript doesn't "initialise" or something?

rkyerevan:
I have the same problem

kjhf:
Seems to be a breakage with NextJS
[...]

PR #671, opened 2021-09-28
#574 handling UNKNOWN ability and allow builds to use Not Set

kjhf:
Addresses #574. 

Also includes a "refresh" button in the analyzer settings to reload the page with the entered abilities, though a neater solution would be to update the URL as changes are made.

Further testing might be needed as I do not have Postgres installed plus some NPM issues.

### DIFF
--- a/components/analyzer/EditableBuilds.tsx
+++ b/components/analyzer/EditableBuilds.tsx
@@ -78,6 +78,10 @@ const EditableBuilds: React.FC<EditableBuildsProps> = ({
     : "SSU";
 
   const isBuildEmpty = (build: Omit<ViewSlotsAbilities, "weapon">) => {
+    // If build is incomplete or malformed, return true.
+    if (build.headAbilities.length !== 4 || build.clothingAbilities.length !== 4 || build.shoesAbilities.length !== 4) {
+      return true
+    }
     return [
       ...build.headAbilities,
       ...build.clothingAbilities,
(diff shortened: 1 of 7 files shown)


Sendouc:
Hey thanks for the PR :)

I need to take another look at this when I'm more energized but just a few quick thoughts. I didn't really write the ticket intending someone to pick it up but it's not a problem in itself. 

One thing that should be noted that these partial builds should not be shown in /builds (probably just gotta change the DB query). Also I'm a bit worried people might submit partial builds by accident since it really is an edge case that they might want to do so in the first place so. Need to think the UX of that. Preferably we'd get a full local environment running with the correct Next.JS version and so on. What were the NPM issues? I will make some issues soon about improving onboarding new contributors so we can discuss those more there.

> Also includes a "refresh" button in the analyzer settings to reload the page with the entered abilities, though a neater solution would be to update the URL as changes are made.

I don't quite understand what is the purpose of this? What problem is it solving?

kjhf:
Welcome! :)

Yeah I did notice when looking through the issues you have a few marked as help-wanted! I hoped that this was a nice starter ticket however and could get something out, quick, while learning the codebase.

The NPM issues were spinning up the instance and general installation on my machine. Was overcome by downgrading as I wrote on #666. I didn't try with any of the DB side.

Honestly I am not fussed if this doesn't get merged in :)

> I don't quite understand what is the purpose of this? What problem is it solving?

When the analyzer is changed, the URL bar/link is not updated which is not intuitive if you come from the builds page, despite the analyzer's stats changing real-time. Therefore, if you copy the URL expecting to share your creation with someone / save as a favourite, you'll have the original build which is probably not desired.

Sendouc:
Hey I made decision not to avoid doing any code changes to the old code base so I will be closing this :(

PR #922, opened 2022-09-06
Translation task (English => Japanese)

aisuzuki:
Added translation json file for Japanese. 
The number of items in each files don't look like sufficient enough, but I could help further translation in case you need more words/sentences to be translated.


Sendouc:
Awesome! Thank you, good job.

Two big parts that are not translatable are the plus server feature (https://beta.sendou.ink/plus/suggestions) and articles (e.g. https://beta.sendou.ink/a/starburst-wins-pax-west). Plus Server is because it has been in history used only by English speakers although that's been changing. Articles are a big effort to translate so that's why I omitted them at least initially.

If you notice some other places that can't currently be translated please open an issue it might be a mistake from me 🤔

Sendouc:
Sorry I didn't consider squashing the pull request will remove your credit and make it look like I made it :( Next time I'll just merge the commits as is so the credit stays. I will make sure to credit you on my Twitter though as having Japanese translations is really awesome.

aisuzuki:
> Sorry I didn't consider squashing the pull request will remove your credit and make it look like I made it

No worry, it would be better for me to hide my real life developer name anyway :) I sent you DM on twitter about my game account name. 
Thank you for the prompt reply and merging, also great commitments for Splatoon community!

PR #929, opened 2022-09-07
Added Korean translation

niLPotential:
I'm still fairly new to Github so I hope I did things right.
Translated the json files to Korean.
I couldn't set the site up locally so I couldn't see it applied.


Sendouc:
Awesome! Good job.

Issue #946, opened 2022-09-14
React error uncontrolled -> controlled input in AbilitiesSelector


Sendouc:
To reproduce

1) Go to http://localhost:5800/u/79237403620945920/builds/new
2) Fill build with any abilities
3) See that once you input the last ability the following error is printed in the console:

![Image](https://user-images.githubusercontent.com/38327916/190124658-851e3f96-6514-4933-986b-36015aa2dac9.png)

PR #949, opened 2022-09-16
Other default value + empty onChange

DoubleCookies:
#946 
Looks like empty array is enough to be controlled after init and be empty so required trigger alert for input.

Empty onChange is because otherwise it will give another error in console (and readonly doensn't work in this case). Maybe it could be done in a better way.
no onchange

### DIFF
--- a/app/routes/u.$identifier/builds/new.tsx
+++ b/app/routes/u.$identifier/builds/new.tsx
@@ -379,8 +379,9 @@ function Abilities() {
         value={
           abilities.flat().every((a) => a !== "UNKNOWN")
             ? JSON.stringify(abilities)
-            : undefined
+            : []
         }
+        onChange={() => {}}
         required
       />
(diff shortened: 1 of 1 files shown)


Sendouc:
I think I forgot a bit context with this since now I remember I put undefined there since the field is marked as "required" and browser will make sure you fill the build completely. With this solution I think they can submit a build without all abilities -> crash. So solution should keep the browser validation for completed build.

Hmmm maybe an empty string would be enough? Or empty string combined with `min/minLength` (i dont remember which) = 1?

DoubleCookies:
I checked it locally and empty array still counts as not filled input field
image

Sendouc (approved):
Oh interesting then I learned something. I changed the placeholder onChange a bit so that linter is happy and added comment :) Check if that's good according to your understanding.

DoubleCookies:
That looks good!

PR #951, opened 2022-09-18
Add remaining DE localization

lolametro:
Not sure why it doesn't detect that the other files/commits were merged already... hopefully merging will still work. If not, I can try to create a new fresh fork.


Sendouc:
Seems to work :D thanks

PR #969, opened 2022-09-25
Adding of Danish Analyser.json translation

FrederikFraJylland:
Analyser.json have been translated to Danish.

If it is possible to get another Danish speaker to review it, it will be advised. Some of the translations are weird to to challenges of translating English video game terms.


Sendouc:
Nice! And yeah it can always be improved later so no problem.

PR #971, opened 2022-09-26
Make `<Ability/>` support keyboard navigation

### LINKED ISSUE #894: <Ability /> lacks keyboard friendliness
Can be given onClick but can't focus since it's just attached to div without tricks. Make it so that this can also be targeted by keyboard when onClick exists or just make it button this case. Otherwise the look and functionality should remain the same.

curtgrimes:
`<Ability/>` can now be tabbed through, and it now shares the same focus outline, hover, and mousedown styles with `<AbilitiesSelector/>`. This includes the `cursor: pointer` style that makes it more obvious that this component is clickable like other buttons.

Fixes #894.

Before:

https://user-images.githubusercontent.com/3038600/192218613-2754506e-520b-4629-b35c-5440afac3b0e.mov

After:

https://user-images.githubusercontent.com/3038600/192218785-8e40da09-d512-47fe-9341-bcee18c213be.mov

### DIFF
--- a/app/components/Ability.tsx
+++ b/app/components/Ability.tsx
@@ -20,7 +20,7 @@ export function Ability({
   const sizeNumber = sizeMap[size];
 
   return (
-    <div
+    <button
       className="build__ability"
       style={
         {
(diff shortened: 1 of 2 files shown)


Sendouc:
Nice one!

PR #975, opened 2022-09-28
Danish Tranlsation

FrederikFraJylland:
Added the missing Danish translation in the front.json file.
Added "buildsGoTo"

Slightly changed translation of  "plus.description":


Sendouc:
👍

Issue #82, opened 2020-09-04
Ability buttons allow dragging


Sendouc:
Currently can only click. Would be intuitive to also drag abilities to slots. Check how it works on https://loadout.ink/en-us/

Builds can be made on user page and build analyzer so this should work on both.

Bonus: Maybe hold Shift to duplicate? So while hold switch and drag from ability already selected it stays in that slot but multiplies and you can drag it to another slot.

PR #979, opened 2022-10-01
Add ability to drag abilities into slots

curtgrimes:
- Do not allow dragging an ability onto a slot that cannot accept that ability.
- When dragging, dim slots that are not a valid target for the ability currently being dragged.
- Do not make any changes to the existing click behavior to assign abilities.

This addresses #82 but I don't know if this closes it: this PR does not implement the bonus idea of holding shift to duplicate abilities across slots. That should be possible, but we would likely have to use a drag/drop library and not necessarily the native drag and drop API (which is what this PR relies on). I think this might add a bit of complexity.

Possible area for further improvement: add the ability to drag a placed ability (one already assigned to a slot) from one slot to another. I can make an issue if this is something you would like.

This does currently only works on desktop. Mobile continues to behave the same as it previously did. (This might be a limitation of using the drag and drop API but I would need to investigate further).

## Demo

Both the build analyzer and the profile builds pages now support drag and drop (blue circles are from the screen recording program and not the webpage):

https://user-images.githubusercontent.com/3038600/193397650-316e498f-5ed5-41b6-bd9a-0f49deb5eb24.mp4

### DIFF
--- a/app/components/AbilitiesSelector.tsx
+++ b/app/components/AbilitiesSelector.tsx
@@ -1,3 +1,5 @@
+import clsx from "clsx";
+import React from "react";
 import invariant from "tiny-invariant";
 import { abilities } from "~/modules/in-game-lists";
 import type { BuildAbilitiesTupleWithUnknown } from "~/modules/in-game-lists/types";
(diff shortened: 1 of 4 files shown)


Sendouc:
Hey this is really good job. I really like the attention to detail with this this UI! A few points:

1) I pushed a commit that gives default values to some Ability props. This is because it was breaking the TS build but check that's it correct.

2) > This addresses https://github.com/Sendouc/sendou.ink/issues/82 but I don't know if this closes it: this PR does not implement the bonus idea of holding shift to duplicate abilities across slots.

I think it closes it. The part about shift key is not that necessary really.

3) > Possible area for further improvement: add the ability to drag a placed ability (one already assigned to a slot) from one slot to another. I can make an issue if this is something you would like.

That sounds like a good enhancement! Go for it.

4) > This does currently only works on desktop. Mobile continues to behave the same as it previously did. (This might be a limitation of using the drag and drop API but I would need to investigate further).

This is a good area to investigate since I think this sort of UI is quite comfortable for mobile. I was checking https://github.com/timruffles/mobile-drag-drop and there it says "Chrome>=96 on Android>=7 and Safari on iOS/iPadOS>=15 are reported to support drag and drop natively" so maybe should be possible?

Sendouc (approved):

Sendouc:
Also I noticed one small issue. Not fixing it yet because of merge conflicts but just putting it here as a note :) https://github.com/Sendouc/sendou.ink/issues/981

You can take a look if you want or otherwise I will get to it later.

I also did a small fix to stop ability buttons triggering form submit https://github.com/Sendouc/sendou.ink/commit/100529d84fe75c2400e6c24e7567d4f003939a60 but I think it should not affect your PR.

curtgrimes:
Thank you! I opened #988 for tracking dragging abilities between slots and opened #989 for mobile dragging.

PR #990, opened 2022-10-04
Fix issue where abilities appeared clickable when they were not

### LINKED ISSUE #981: Ability appears clickable even if it is not (BuildCard)
BuildCard abilities appear clickable:

https://user-images.githubusercontent.com/38327916/193400747-a6fac647-de86-4ff5-815a-64da077ebf2c.mov

Maybe: have Ability take prop ("presentational" or whatever) that decides whether to render a button or a div?

curtgrimes:
Fix issue where abilities appeared clickable when they were not by adding a `readonly` attribute set on for instances of `<Ability/>` in `<BuildCard/>`.

Issue was introduced in #971.

Fixes #981

https://user-images.githubusercontent.com/3038600/193729074-36cbe7da-c34e-446f-9986-8e6a3cdb8143.mov

### DIFF
--- a/app/components/Ability.tsx
+++ b/app/components/Ability.tsx
@@ -13,6 +13,7 @@ const sizeMap = {
 export function Ability({
   ability,
   size,
+  readonly = false,
   dragStarted = false,
   dropAllowed = false,
   onClick,
@@ -20,6 +21,7 @@ export function Ability({
 }: {
   ability: AbilityWithUnknown;
   size: keyof typeof sizeMap;
+  readonly?: boolean;
   dragStarted?: boolean;
   dropAllowed?: boolean;
   onClick?: () => void;
@@ -44,6 +46,7 @@ export function Ability({
         "is-drag-target": isDragTarget,
         "drag-started": dragStarted,
         "drop-allowed": dropAllowed,
+        readonly,
       })}
       style={
         {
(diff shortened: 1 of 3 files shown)


app/components/Ability.tsx
```diff
@@ -58,7 +61,7 @@ export function Ability({
         setIsDragTarget(false);
         onDrop?.(event);
       }}
-      type="button"
```
Sendouc, line 61: https://user-images.githubusercontent.com/38327916/193895399-51aa9b88-9fa4-40aa-995b-c22a6006cfc2.mov

Is it important to remove this attribute? Maybe only conditionally reading readonly? See video for the bug it will cause when adding new build (tries to submit new build on ability click)

Sendouc (commented):
Check comment but otherwise I like this solution 💯

app/components/Ability.tsx
curtgrimes: Sorry, I inadvertently removed `type="button"`. Should absolutely stay in. I'll push a fix. Thanks!

curtgrimes: Fixed in https://github.com/Sendouc/sendou.ink/pull/990/commits/36deb0c58820edf464cb35f352c911ee402e6f27

PR #991, opened 2022-10-04
Fix incorrect StatCard title alignment

### LINKED ISSUE #987: StatCard titles not aligned
image

Without ability StatCard visuals bugged. Base / Build titles should be on same line for every card ability or no (empty space at the bottom if no ability for that StatCard).

curtgrimes:
When a StatCard displays an ability, the title is incorrectly positioned vertically at the bottom of the StatCard. After this change, it is aligned near the top.

Fixes #987

## Before

CleanShot 2022-10-03 at 23 09 49@2x

## After

CleanShot 2022-10-03 at 23 09 36@2x

### DIFF
--- a/app/styles/analyzer.css
+++ b/app/styles/analyzer.css
@@ -121,6 +121,7 @@
   display: flex;
   justify-content: space-evenly;
   margin-top: var(--s-2);
+  margin-bottom: auto;
   gap: var(--s-1);
 }
 


Sendouc:
I think maybe it should still keep the stat aligned on same level like in prod:

image

vs. prod:

image

I think it would be a cleaner design 🤔 idea: what if when ability is missing we just render some placeholder ability and apply `visibility: hidden` to it? And then keep it aligned to bottom. Or maybe there is a better way to do the same thing (maybe apply fixed height to the ability container) and have the same effect.

halp-pat:
You can get it aligned with flex box and an empty element. See https://github.com/Sendouc/sendou.ink/pull/997

Sendouc:
Yeah I could have written the issue a bit better..

curtgrimes:
> Yeah I could have written the issue a bit better..

No worries! #997 looks great.

PR #992, opened 2022-10-04
Select dropdown arrow position should use fixed value instead of percent

### LINKED ISSUE #983: Select dropdown arrow spacing not consistent
Screenshot 2022-10-02 at 11 33 56

Padding to the right of arrow depends on select width. Instead should always be the same.

halp-pat:
fixes #983


Sendouc:
Nice one. I was so annoyed by this 😅

PR #997, opened 2022-10-05
Stat Card values should be on the same line

### LINKED ISSUE #987: StatCard titles not aligned

halp-pat:
fixes #987


Sendouc:
Nice solution 👍

PR #998, opened 2022-10-05
Use fixed widths for BuildCard

### LINKED ISSUE #993: BuildCard overflow issue on mobile
image

image

halp-pat:
fixes #993

Seems like it's caused by the browser's default font size. The ability card gets too small for the fixed width images.


Sendouc:
💯

PR #1010, opened 2022-10-13
Add participant count in user page results table

### LINKED ISSUE #864: Participant count in user page results table
New column that displays the participant count of the event.

remmycat:
Closes #864 

Notes:
- ~~The `check-translation-jsons` script already found changes on the current head, which is why there are two commits with changes to `translation-progress.md`.~~ Fixed after rebase
- Since the table became wider with this change I checked some smaller viewports and noticed that the scroll on overflow was broken for the table, so I fixed this by making sure the `main` element is normally just as wide as the parent.

### DIFF
--- a/app/routes/u.$identifier/results.tsx
+++ b/app/routes/u.$identifier/results.tsx
@@ -15,7 +15,7 @@ export default function UserResultsPage() {
   const data = parentRoute.data as UserPageLoaderData;
 
   return (
-    <main className="u__results-main layout__main">
+    <main className="main layout__main">
       <Section className="u__results-section">
         <table>
           <thead>
--- a/app/routes/u.$identifier/results.tsx
+++ b/app/routes/u.$identifier/results.tsx
@@ -23,6 +23,7 @@ export default function UserResultsPage() {
               <th>{t("results.placing")}</th>
               <th>{t("results.team")}</th>
               <th>{t("results.tournament")}</th>
+              <th>{t("results.participants")}</th>
               <th>{t("results.date")}</th>
               <th>{t("results.mates")}</th>
             </tr>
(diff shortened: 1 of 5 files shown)


app/styles/u.css
```diff
@@ -147,11 +147,6 @@
   font-weight: var(--bold);
 }
 
-.u__results-main {
-  max-width: 52rem;
-  margin: 0 auto;
-}
```
remmycat, line 153: My first idea was to change this to:

```css
My first idea was to change this to:

```css
.u__results-main {
   width: 100%;
   max-width: 52rem;
   margin: 0 auto;
 }
```

But then I realised that the user profile page just uses the `.main` class defined in `layout.css`, which has this line and also a padding that made the table's div not touch the window edges on small viewports.

One difference is that `.main` uses `48rem` as a max-width (including the padding i believe).
Let me know if that class shouldn't be used here / If the 52rem were purposefully picked.

remmycat:
I tried rebasing onto the current state of the `rewrite` branch and got rid of the extra commit for the translation progress.
I'm not sure why the GitHub UI says there are still conflicts?

app/routes/u.$identifier/results.tsx
```diff
@@ -37,6 +38,7 @@ export default function UserResultsPage() {
                     {result.eventName}
                   </Link>
                 </td>
+                <td>{result.participantCount}</td>
```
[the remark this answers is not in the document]
Sendouc: I think showing nothing makes perfect sense!

app/styles/u.css
Sendouc: I don't remember if there was some special meaning but just testing your PR this looks great so let's go with this 👍

Sendouc:
Veryy nice

PR #1012, opened 2022-10-14
Deactivate i18next interpolation escaping

### LINKED ISSUE #994: Special characters not shown correctly in delete build popup
Why shows correctly in build title but not delete build modal?

Example character: `/`

image

image

remmycat:
As talked about in the ticket, the interpolation escaping i18next provides is not beneficial in combination with React in this case.

Fixes #994


Sendouc:
Yeah really nice 👍

PR #1015, opened 2022-10-15
Improve result placement display with custom medal icons

### LINKED ISSUE #1008: Tournament result icons could be more readable
Opened someone's profile today and really didn't know if that was gold or bronze, the icon is very small in desktop and there's no text on hover. Just a small adjustment i think could be improved

image

remmycat:
Closes #1008 

I also tried to improve the placement ordinals a bit 

4th, 23th, 11th
should now show up as
4<sup>th</sup> 23<sup>rd</sup> 11<sup>th</sup>


remmycat:
I'm a little unsure still about the whole "english ordinals for every language" thing and if the comment is true that it is used internationally. 

In theory I only added onto how it worked before where it always added the "th", but maybe we should just remove the suffix for non-english languages?

Sendouc:
Very nice job again :)

Good point about the ordinals that was one place where I forgot to add support for translating. Made a follow up issue: https://github.com/Sendouc/sendou.ink/issues/1016

PR #1017, opened 2022-10-16
Fix edit badge dialog flickering

### LINKED ISSUE #826: Loading edit badges modal flickers when SSR
https://user-images.githubusercontent.com/38327916/177769213-915c2afd-26ae-47cd-a1b2-deaf5e43c20d.mov

See wrong behaviour with /badge/$id/edit modal and correct behaviour with Plus Server suggestion modal. What is the difference? Fix

remmycat:
Closes #826

To be honest, I'm not a 100% sure how the CSS caused the glitched behaviour.
However, the reason why this modal was behaving different than the others, is that the "stack md" classes were applied to the dialog itself, instead of the container inside.
I assume the blame then goes to the css overwriting the dialog's `display: block` with `display: flex`.

### DIFF
--- a/app/routes/badges/$id/edit.tsx
+++ b/app/routes/badges/$id/edit.tsx
@@ -72,8 +72,8 @@ export default function EditBadgePage() {
   const { badgeName } = useOutletContext<BadgeDetailsContext>();
 
   return (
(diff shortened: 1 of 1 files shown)


Sendouc:
Nice find!

PR #1020, opened 2022-10-17
User result highlights

### LINKED ISSUE #1019: Allow users to highlight results
When users have a lot of results they might want to show their best results first. For this case let's allow users to pick any number of results that shows first before other results.

- implement new action on user results page.
- form UI can be for example toggles per event name

UI can be something like:

```
> Highlighted results
Table with highlighted results ordered by date 

> Other results
Table with non-highlighted results ordered by date 
```

Where highlighted result is a title and other results is a title or maybe accordion?

## Current situation:

image

remmycat:
Closes #1019 

Let me know what you think of the UI and if I can improve anything 🙂


Sendouc:
I checked UI and I really like your implementation 👀 👀 👀 👀 

I will make sure to review this fully as soon as I have a good moment during this week!

remmycat:
I'm glad!

And no worries if it takes some time for you to have that moment, there's no rush from my side

remmycat:
Force-pushed this to rebase, remove a leftover comment, and adjust the css of the highlight table a little bit (remove the inline padding when we have the highlight border)

### Screenshots of the current implementation

**UI on own profile's results page:**
Screenshot 2022-10-18 at 09 38 29

**UI when clicking on "choose highlights":**
Screenshot 2022-10-18 at 09 43 52

**After saving:**
Screenshot 2022-10-18 at 09 38 42

**How it appears to other users**
Screenshot 2022-10-18 at 09 39 25

app/utils/arrays.ts
```diff
@@ -34,3 +34,9 @@ export function joinListToNaturalString(arg: string[]) {
 
   return last ? `${commaJoined} and ${last}` : commaJoined;
 }
+
+export function normalizeFormFieldArray(
+  value: undefined | null | string | string[]
```
Sendouc, line 39:
```suggestion
  value: unknown
```

Maybe?

migrations/011-user-result-highlights.js
```diff
@@ -0,0 +1,17 @@
+module.exports.up = function (db) {
+  db.prepare(
+    `
+    create table "UserResultHighlight" (
+      "teamId" integer not null,
+      "userId" integer not null,
+      foreign key ("teamId") references "CalendarEventResultTeam"("id") on delete cascade,
+      foreign key ("userId") references "User"("id") on delete cascade,
```
Sendouc, line 8: Should add index on these FK's. On some migrations I forgot it... `007-builds.js` is an example of one that has them.

Sendouc (changes requested):
Looks really good 💯 Both design and code! Only thing that in my opinion should be changed is adding the indexes for performance reasons but otherwise this is good to go :) Other comments just for consideration.

app/db/models/calendar/findResultsByUserId.sql
```diff
@@ -12,7 +12,16 @@ select
...
+  ) as "startTime",
+  (
+    select
+      1
+    from
+      "UserResultHighlight"
+    where
+      "userId" = @userId and
+      "teamId" = "CalendarEventResultTeam"."id"
+  ) as "isHighlight"
```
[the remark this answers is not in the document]
remmycat: Huh, yeah! Gonna change it to the exists way for consistency

migrations/011-user-result-highlights.js
remmycat: Will add them! I'm not that well-versed in SQL so I'm flailing a lot and am usually already happy if things work :D

app/utils/arrays.ts
remmycat: Hmm, it's quite specifically one of those values the browser sends with the form when a specific input name is repeated (though maybe not null). When typing it as unknown I would have to check the array children for being strings and handle errors and so on.

remmycat:
Comments should be handled 🙂

PR #1022, opened 2022-10-18
Show dialog for failed login attempts

### LINKED ISSUE #824: Auth when clicking cancel it leads to an error page
Currently when you click "Log in" but on discord.com click "Cancel" it leads you to this kind of page:

image

Instead it should leads us back to sendou.ink UI with some helpful message that says you should approve the log in on Discord.

remmycat:
Closes #824 

Screenshot 2022-10-18 at 23 19 44
Screenshot 2022-10-18 at 23 20 25

### DIFF
--- a/app/components/layout/UserItem.tsx
+++ b/app/components/layout/UserItem.tsx
@@ -1,9 +1,10 @@
-import { Link } from "@remix-run/react";
+import { Link, useSearchParams } from "@remix-run/react";
 import { useTranslation } from "react-i18next";
 import { useUser } from "~/modules/auth";
 import { LOG_IN_URL, LOG_OUT_URL, userPage } from "~/utils/urls";
 import { Avatar } from "../Avatar";
 import { Button } from "../Button";
+import { Dialog } from "../Dialog";
 import { DiscordIcon } from "../icons/Discord";
 import { LogOutIcon } from "../icons/LogOut";
 import { UserIcon } from "../icons/User";
@@ -12,8 +13,9 @@ import { Popover } from "../Popover";
 export function UserItem() {
   const { t } = useTranslation();
   const user = useUser();
+  const [searchParams, setSearchParams] = useSearchParams();
 
-  if (user)
+  if (user) {
     return (
       <Popover
         buttonChildren={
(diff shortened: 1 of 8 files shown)


remmycat:
Not super sure about the wording here still.

- Should the "cancelled" text specify that connections are only shown on a profile if they are configured as public on discord? How to phrase this without blowing up the dialog?
- Should the "unknown" case provide a specific link or linkbutton for getting help, like the sendou.ink discord link?

Sendouc (commented):
Very nice solution again! Looks great as well. Left a few small comments for consideration but otherwise looks good to be merged.

Regarding the text I think what you wrote is perfectly fine. Maybe could just say "social connections" instead of listing out the different services? But I don't think it's a big deal either way. Most important thing is that its clear and I think it definitely is.

app/components/layout/UserItem.tsx
```diff
@@ -45,16 +47,55 @@ export function UserItem() {
...
+      <form action={LOG_IN_URL} method="post" data-cy="log-in-form">
+        <button
+          type="submit"
+          className="layout__log-in-button"
+          data-cy="log-in-button"
+        >
+          <DiscordIcon /> {t("header.login")}
+        </button>
+      </form>
+      {authError != null && (
```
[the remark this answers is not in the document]
remmycat: Oh, I get that, but this is actually the only case where I purposefully do write `==`/`!=`, because `== null` is specifically true for undefined and null, but nothing else.

So in my code style I usually use `== null` as a shorthand for just checking "is this null or undefined", without having to care about finding out which specific variant is relevant in this case, nor having to deal with special cases for 0 or "".

It's basically the same check the nullish coalescing operator (`??`) uses

But all that aside, I can definitely change this to be more specific, I just wanted to explain where I'm coming from :)

public/locales/en/user.json
```diff
@@ -20,5 +20,10 @@
   "forms.errors.invalidCustomUrl.numbers": "Custom URL can't only contain numbers",
   "forms.errors.invalidCustomUrl.strangeCharacter": "Custom URL can't contain special characters",
   "forms.errors.invalidCustomUrl.duplicate": "Someone is already using this custom URL",
-  "forms.errors.invalidSens": "Motion sens can't be set if R-stick sens isn't"
+  "forms.errors.invalidSens": "Motion sens can't be set if R-stick sens isn't",
+
+  "auth.errors.cancelled": "Login Cancelled",
```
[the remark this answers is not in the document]
remmycat: I always get tricked by that 😅
I went with "aborted" now. Thinking about it, it seemed more fitting to me, as the login process already started from the website's point of view.

app/components/layout/UserItem.tsx
```diff
@@ -45,16 +47,55 @@ export function UserItem() {
         </div>
       </Popover>
     );
+  }
+
+  const authError = searchParams.get("authError");
+  const closeAuthErrorDialog = () => {
+    searchParams.delete("authError");
+    setSearchParams(searchParams);
```
Sendouc, line 55: I believe I missed this first time but I think this should be changed: https://beta.reactjs.org/learn/updating-objects-in-state#treat-state-as-read-only

`searchParams` in this case is state so we should treat it as readonly. Options: first make copy of it before modifying or see if you can give object to setSearchParams as argument or see if you can make functional update work.

Sendouc (commented):
Just one last thing I spotted! Otherwise 💯 💎

public/locales/en/user.json
```diff
@@ -25,5 +25,10 @@
   "forms.errors.invalidCustomUrl.numbers": "Custom URL can't only contain numbers",
   "forms.errors.invalidCustomUrl.strangeCharacter": "Custom URL can't contain special characters",
   "forms.errors.invalidCustomUrl.duplicate": "Someone is already using this custom URL",
-  "forms.errors.invalidSens": "Motion sens can't be set if R-stick sens isn't"
+  "forms.errors.invalidSens": "Motion sens can't be set if R-stick sens isn't",
+
+  "auth.errors.aborted": "Login Aborted",
+  "auth.errors.failed": "Login Failed",
+  "auth.errors.discordPermissions": "For your sendou.ink profile, the site needs access to your Discord profile's name, avatar and social connections.",
+  "auth.errors.unknown": "The login via Discord failed for an unknown reason. If this keeps happening, please reach out for help."
```
remmycat, line 33: I realised that these keys have to be part of the common translations as they can be shown on any page the way it is currently set up, and we don't really want to load the `user` translations for all pages.

I didn't get how remix-i18next worked until just now 😅

Sendouc: Ah! Very true. Yeah I guess you got it but we load what translations each page has on the handle function :)

```
export const handle = {
  i18n: "calendar",
};
```

app/components/layout/UserItem.tsx
remmycat: Yeah, I was being lazy, will update!

remmycat:
Should all be done now :)

PR #1023, opened 2022-10-19
Added missing Danish translations.

FrederikFraJylland:
Added Danish translations for the map list and profile page.

Note.

The Danish profile page translations need to be reviwed, as some of the translations might be too long to fit in thier "boxes".


Sendouc:
👍 

Quickly checked profile and lgtm

PR #1027, opened 2022-10-20
Small css improvements

remmycat:
The mode selection was super glitchy with safari, it would decide at seemingly random times when to (not) apply the grayscale filter, sometimes only when its's selected, sometimes only when not.


Sendouc:
👍

PR #1028, opened 2022-10-20
Improve check-translations-jsons speed

remmycat:
Script execution took 17 seconds for me before, now it's 2.

I also prefer if the script doesn't implicitly format all files, which I might be currently editing 😅


Sendouc:
Haha yeah I was too lazy with this but your solution is much nicer.

PR #1029, opened 2022-10-20
Generic Breadcrumbs component/mechanism

remmycat:
I saw this in the Remix docs as an example for using the handle export and thought this would probably be cool to implement :)

It's probably easiest to look at the two commits separately

### DIFF
--- a/app/utils/remix.ts
+++ b/app/utils/remix.ts
@@ -1,4 +1,6 @@
 import { z } from "zod";
+import { type ReactNode } from "react";
+import { type Namespace } from "react-i18next";
 
 export function notFoundIfFalsy<T>(value: T | null | undefined): T {
   if (!value) throw new Response(null, { status: 404 });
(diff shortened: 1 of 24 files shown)


app/utils/remix.ts
```diff
@@ -89,3 +91,15 @@ export function validate(condition: any, status = 400): asserts condition {
...
+ * Our custom type for route handles - the keys are defined by us or
+ * libraries that parse them.
+ *
+ * Can be set per route using `export const handle: SendouRouteHandle = { };`
+ * Can be accessed for all currently active routes via the `useMatches()` hook.
+ */
+export type SendouRouteHandle = {
+  /** The i18n translation files used for this route, via remix-i18next */
+  i18n?: Namespace;
+};
```
Sendouc, line 113: Woo I love this.

Sendouc:
Neat B)

PR #1034, opened 2022-10-22
Allow users to copy a map pool of another event

### LINKED ISSUE #1013: New calendar event form dropdown to insta-select pool of your previous tournament
image

remmycat:
Closes #1013 

I went a bit back and forth on how to write the data loading part. My thoughts:

- A backend full text search is probably the most scaling, but it seemed like a bit much to set up, and I have no experience with that. From what I found we would use the sqlite FTS 5 extension?
- Copying the current users approach of loading them all and keeping them in a cache seemed fine for the event map pools, but it would be weird if you add multiple events and cannot reference the one you added before. (At least I think that would be the case with the swr immutable? And re-fetching everything in a timed manner seemed also weird
- Adding a high limit (100-200?) to the sql seemed fine-ish but It would be kinda weird that you would suddenly not be able to not find some older events, if you always used them as reference

In the end I went for a compromise for now with loading all the events first as part of the route loader, but keeping the data as minimal as I could, only including the event ID and name, the actual map pool only gets fetched on-demand when it is selected, which is pretty fast.

With the UI I tried to use the blue info color to convey the connection between the combobox and which stages will be activated by the template.

### DIFF
--- a/app/components/Combobox.tsx
+++ b/app/components/Combobox.tsx
@@ -3,7 +3,7 @@ import * as React from "react";
 import Fuse from "fuse.js";
 import clsx from "clsx";
 import type { Unpacked } from "~/utils/types";
-import type { GearType, UserWithPlusTier } from "~/db/types";
+import type { CalendarEvent, GearType, UserWithPlusTier } from "~/db/types";
 import { useUsers } from "~/hooks/swr";
 import { useTranslation } from "react-i18next";
 import {
@@ -36,6 +36,7 @@ interface ComboboxProps<T> {
   initialValue?: ComboboxOption<T>;
   clearsInputOnFocus?: boolean;
   onChange?: (selectedOption?: ComboboxOption<T>) => void;
+  showEmptyQueryResults?: boolean;
 }
 
 export function Combobox<T extends Record<string, string | null | number>>({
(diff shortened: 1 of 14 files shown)


remmycat:
Accidentally marked it as a draft first, this is ready

remmycat:
Just realised this should probably also be available on the normal maps page, now trying to find out if there is a difference between the different `MapPoolSelector` components and if i can unify them 

--> Back to WIP

remmycat:
So it seems like the MapPoolSelector components were basically the same, only difference was that one was able to be set to presentational and the other one wasn't. So I'm gonna unify them.

Sendouc:
Yeah with `MapPoolSelector` I think I just forgot to unify them in the end... So great if you can do it 😅 

This is looking great! Will be really useful to people who submit a lot of tournaments. Just wondering about UX. It was a bit unclear for me how it works here:

image

I think user might expect that the blue ones are selected but in fact that ring is just used for reference? So I wonder if simpler solution where it just selects the exact map pool would be more clear. I think the point with the blue ring is smart but yeah I'd still preselect those since at least I'm thinking that's how users might expect it to work.

remmycat:
Yeah, that suggestion makes a lot of sense to me! I didn't realise that I only had the issue of having to show the user which fields are getting affected, because it isn't getting applied immediately 🤦🏻

remmycat:
Closed in favour of #1047

PR #1047, opened 2022-10-26
Map pool templates

### LINKED ISSUE #1013: New calendar event form dropdown to insta-select pool of your previous tournament

remmycat:
Closes #1013

This ended up being quite a chonky PR because I ended up reworking lots of related areas a little.
But I put lots of effort into making the commits clean increments, as if I had a clear vision of where I wanna end up all along 😁 

So it's probably nice to review this commit by commit!
If you feel like it, you can also check them out individually, each one should be in a working state.

Notes: 
- The plus icon was made by me, low-effort editing the Cross icon svg. It's probably not perfectly consistent with the others. I have no idea if there is an icon set we would usually get these from, or if you wanna pick your own.
I'm also fine with using different icons or maybe button labels there if you can think of something good.
- The data loading now ended up being:
    - I ended up fetching all events for the combobox in the same way you fetched the users, in-memory-caching them with swr.
    - The routes that show the editor additionally load the 5 most recent events in their loader and show them in the dropdown - this should allow even just-created events to show up there.
    - I ended up sending the map pools always in serialised form to the user to keep the response as small as possible
- I tried fiddling around a bit with the colours and filter of the mode toggles, because there were some contrast issues, especially in light mode. No worries if you don't like it, happy to get feedback :)
- Unsure if i should switch name of "Custom" template to "None", since you don't actually have a template in that case 🤔 
- The selects were weird for the layout because they would consistently overflow their parent container, I hope the change from width 90% to 100% is fine. (90% would overflow because the box-sizing property that was set for all elements was unset at the start of the css block)

### DIFF
--- a/app/components/Combobox.tsx
+++ b/app/components/Combobox.tsx
@@ -4,7 +4,7 @@ import Fuse from "fuse.js";
 import clsx from "clsx";
 import type { Unpacked } from "~/utils/types";
 import type { GearType, UserWithPlusTier } from "~/db/types";
-import { useUsers } from "~/hooks/swr";
+import { useAllEventsWithMapPools, useUsers } from "~/hooks/swr";
 import { useTranslation } from "react-i18next";
 import {
   clothesGearIds,
@@ -15,6 +15,7 @@ import {
 } from "~/modules/in-game-lists";
 import { gearImageUrl, mainWeaponImageUrl } from "~/utils/urls";
 import { Image } from "./Image";
+import { type SerializedMapPoolEvent } from "~/routes/calendar/map-pool-events";
 
 const MAX_RESULTS_SHOWN = 6;
 
@@ -36,6 +37,7 @@ interface ComboboxProps<T> {
   initialValue?: ComboboxOption<T>;
   clearsInputOnFocus?: boolean;
   onChange?: (selectedOption?: ComboboxOption<T>) => void;
+  fullWidth?: boolean;
 }
 
 export function Combobox<T extends Record<string, string | null | number>>({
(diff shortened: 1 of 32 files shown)


Sendouc:
So for some funny reason reviewing the code commit by commit got both my Chrome and Firefox to crash..? 😅 I'm thinking it could be a GitHub bug so I'll check back later to review the code. 

But other first thoughts.... I love this 💯 🍾 🎆  Really great job! Very useful feature and implemented beautifully. I could see this being very useful with the upcoming "TO Tools" team map selection as well. Some random thoughts from trying the feature out:

- I think as you said None might be better default name than Custom
- "Only Rain Maker" -> "Only Rainmaker"
- Icon set I use is Heroicons but anything like it is fine as well :)

Some usability questions we could go over:

image

I'm thinking colors here could be switched so that Clear is red as to me it's more destructive than Remove. Even if you Remove if you add again the pool is where you left it but not like that with Clear.

image

Small thing but there is uneven gap between "Map Pool" label and button compared to other form items.

image

I'm still wondering about this. With "+" I was somehow first thinking I need to add the maps stage by stage. So I was thinking is there some more descriptive icon for "select all". Or just short text "All"? Maybe it's not a problem just something I thought for a second.

remmycat:
Thanks for the feedback! 

- Now that you mention it, showing Clear as more destructive makes sense to me!
- And yeah same, with the icon 😩 One thing I worry about with the text label is the layout and preventing layout shift. I did make a bit more room on the pages if we're not on mobile, but "All" will likely not be 3 characters in every language.
- I'll have a look at HeroIcons, either for the plus, or for finding a different icon! The description should maybe also change from "Add" to "Select All"

remmycat:
And yeah, sorry for making your browser crash, for me everything showed up fine 😅

remmycat:
I kinda like `arrow-long-left`. I also prefer if the select all and clear icons don't look so similar, after clicking them for a while my mind kinda glitched out with all the icons looking the same but at different angles 🥴
Not sure if tiny or not is better:

Screenshot 2022-10-27 at 09 14 19
Screenshot 2022-10-27 at 09 15 12

remmycat:
All the feedback so far should be applied 🙂 Went with the tiny icons for now.

remmycat:
> My reasoning is that while the places you used it seemed fine my impression is that it's easy for people to suppress the linter rule when it creates actual bugs.

My thinking was that if you call a hook named `useOnce()`, it's not there to memoize anything (and introduce bugs with stale data), but to be evaluated exactly once per component instance. If that creates a bug, maybe you should have called `useTwice()` instead 😛 

Maybe I could rename it to be much more explicit and reword the comment to explain what it does, instead of making it sound like it's there for suppressing linter warnings?

But I can also definitely remove it if you think people might call this accidentally when they actually should call `useMemo` and see warnings.

Sendouc:
I think that's fair point :) Let's go with it!

Issue #1049, opened 2022-10-26
Theme choice does not persist when the browser session is terminated


ElementUser:
Theme choice does not persist when the web browser session is terminated.

I'm a dark mode user for the website & I get a bit irritated when I have to retoggle the theme every time I visit the site on a different day.

I will be submitting a fix for this in the near future (just need to clean up my local branch).

P.S. Thank you for the awesome website :)

ElementUser:
Added my PR here that addresses this issue: https://github.com/Sendouc/sendou.ink/pull/1050

Sendouc:
I think https://github.com/Sendouc/sendou.ink/commit/b340f32ae38e1487cf73818a631ad27ea12e6b39 fixes it

Can you test when you are able?

ElementUser:
> I think [b340f32](https://github.com/Sendouc/sendou.ink/commit/b340f32ae38e1487cf73818a631ad27ea12e6b39) fixes it
> 
> Can you test when you are able?

@Sendouc I checked out that commit & ran the app locally, I can confirm that your commit fixes the issue!! :)

I will close my PR - thanks again for fixing it! :)

PR #1050, opened 2022-10-27
Bug/persist user theme choice

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1049

# Description of Changes

Theme choice from the user now persists via localStorage in that user's web browser.
- In other words, the user's theme choice persists even after ending that web browser's session and opening a brand new browser session.

Also updated the serverside-rendering workaround for the theme.
- Note that there was some unavoidable hardcoding that was needed here, as the workaround was already messy with some raw JS injection into the HTML element.

Also added a suite of Cypress E2E tests for the Theme Switcher.

## Notes

I've read about some kind of a flash issue with the wrong theme. I was not able to replicate this on my end, so please ensure that this theme flash issue does not occur. I will point this out in the code as well.

## Manual testing steps

1. Check out the latest version of this branch, run the application locally with `npm run dev`, visit the website ran locally (default URL when hosting locally: http://localhost:5800/), then click on the Theme Switch button.
2. Close the web browser, and make sure to **close all windows of the same type of web browser**. Example: if you visited the website on Google Chrome, you must close all Google Chrome instances.
3. Re-open a brand new instance of the same web browser, then re-visit the website ran locally.
4. Expectation: you should see that your selected theme persists.

If you do these exact same steps with the currently live version of the website (https://sendou.ink/), then you will see that the theme does not persist properly.

## Cypress test results when running them on my end:
image

## Browsers this was tested on:
- Google Chrome
- Firefox
- Microsoft Edge

### DIFF
--- a/app/modules/theme/provider.tsx
+++ b/app/modules/theme/provider.tsx
@@ -2,6 +2,8 @@ import { useFetcher } from "@remix-run/react";
 import type { Dispatch, ReactNode, SetStateAction } from "react";
 import { createContext, useContext, useEffect, useRef, useState } from "react";
 
+const THEME_LOCAL_STORAGE_KEY = "theme-preference";
+
 enum Theme {
   DARK = "dark",
   LIGHT = "light",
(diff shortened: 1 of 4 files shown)


app/modules/theme/provider.tsx
```diff
@@ -85,14 +101,20 @@ function ThemeProvider({
   );
 }
 
-// this is how I make certain we avoid a flash of the wrong theme. If you select
-// a theme, then I'll know what you want in the future and you'll not see this
-// script anymore.
+/** 
+ * This is how I make certain we avoid a flash of the wrong theme.
```
ElementUser, line 105: Note: I was not able to replicate this on my end, so please ensure that this theme flash issue does not occur. Thank you!

Sendouc:
Hey there! Thanks for the PR :) Even adding tests that's very nice 👍  

How could I reproduce this issue? For me it works so that it remembers I prefer Dark mode as it's stored in the cookie. Do you have cookies disabled? I would need to dig in how it affects it to read it from local storage vs. cookie. There are some extra considerations with SSR.

ElementUser:
@Sendouc awesome!

Yeah I didn't realize that it's possible something else broke & that the theme was stored via a cookie.

The issue in that case is that for whatever reason, the cookie for the theme gets deleted after closing all web browser sessions. Maybe a better fix would be to address that bug instead. Either way though, just let me know & maybe I can even look into the cookie persistence issue when I have time

ElementUser:
Closing due to the original issue actually being cookie-related. This is the proper fix from Sendou: https://github.com/Sendouc/sendou.ink/commit/b340f32ae38e1487cf73818a631ad27ea12e6b39

PR #1054, opened 2022-10-27
Update translation  (JA)

aisuzuki:
Sorry but it's incomplete pull request.
I tried to translate remaining items, but some of them doesn't make sense without context. Is there a staging server to check if translations are right and natural? 
e.g,)  stat.special.deviceHp, which device does it mean?

not-yet-translated list)
   stat.special.deviceHp
  "stat.sub.velocity":
  "stat.sub.firstPhaseDuration": 
  "stat.sub.secondPhaseDuration
  "damage.MAX_CHARGE":
  "consumptionExplanation"
  "maps.description"


Sendouc:
Hey!

> stat.special.deviceHp
Big Bubbler the weak point so the thing on the top and inside the bubble.

"stat.sub.firstPhaseDuration":
"stat.sub.secondPhaseDuration
These are about Sprinkler. It has different phases of power. First phase outputs the most ink but is limited in time how long it lasts.

If you have questions about any other translation strings feel free to ask. Sadly there is no staging server so only way to easily see the translations in action would be to set the project up following instructions in the README.

aisuzuki:
OK thank you! I translated rest of items by looking prod website :) 
also I brushed up some translations + added new items from last commit in master.

It would be great if there's a way to let users send feedback about translation, too. 
I guess this PR is ready to be merged.

Sendouc:
Nice job :) Currently the idea is that users can use our Discord for feedback but it's true that it might not be that convenient always.

Issue #1036, opened 2022-10-24
Analyzer ability icons appear clickable


Sendouc:
They are not buttons so should not have that cursor + active effect:

https://user-images.githubusercontent.com/38327916/197572886-89abbda0-9510-4ba1-9fb4-7be6b4346cad.mov

remmycat:
Probably the same component anyway, but the same goes for the icons in the dmg calc analyzer 🙂

ElementUser:
I submitted a PR here: https://github.com/Sendouc/sendou.ink/pull/1057

PR #1057, opened 2022-10-28
Issue 1036: non-draggable ability icons will no longer have its curso…

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1036

# Description of Changes

Non-draggable ability icons will no longer have its cursor morphed to a pointing hand icon. 

Some detailed changes include:
- Ability icons are now treated as `readonly` if no onClick function is passed into them
- Conditionally rendering the Abiliity icons as a `<div />` or `<button />` element depending on if they are supposed to have the `readonly` class or not
- Removed unnecessary tabIndex adjustment for AbilityTag element if it is `readonly` (since it is no longer rendered as a button if it is readonly)
- We now also condtionally give type="button" to AbilityTag only if it is actually a button

Thanks to @remmycat for the suggestions!

### Build Analyzer page

All ability icons that aren't `?` icons will still have its cursor morphed to a pointing hand icon upon hovering over them (which I believe is still the intended/correct behaviour): 

image

- Note that the Tacticooler special icon does not do this (since you aren't able to drag it)
- Same behaviour applies to the sub-weapon icon and special icon for the selected weapon (Curling Bomb and Ultra Stamp in this example)

The ability icons on the right half of the page are not draggable, and thus their mouse cursor will not change when hovering over them:

image

### Object DMG Calc page

As mentioned by @remmycat  in the original issue, this bug is also fixed for the ability icons on the Object DMG Calc page:
image

## Tab Index testing

I made sure I couldn't tab to visible readonly Ability icons by pressing Tab repeatedly until I cycled through all of the visible Ability icons on the page & confirm that this is indeed the case.


Sendouc:
This is a neat solution for the cursor problem! :) One more thing that should still be done is making sure they don't get the active style of button: https://github.com/Sendouc/sendou.ink/blob/rewrite/app/styles/common.css#L53

Testing your PR this still seems to happen when I press on them even if the cursor is now correct.

remmycat:
Yeah, thanks for having a look at this issue @ElementUser!

I found 2 things we should keep in mind for this fix:

1. There is already a `.readonly` class for the abilities, that applies the same style you added.<br/>I believe it might make sense to not change the css rule for all ability components. Instead we could set the `<Ability readonly />` prop for non-interactive abilities, which then applies the `.readonly` css class 🙂 
2. This would definitely be a bonus, but no worries if you wanna keep it out of this PR: For web accessibility (e.g. accessing the site as a blind user with a screen reader), it's very important to use the correct HTML elements semantically. If the <Ability> is not a button, it should definitely not be rendered as a `<button>`.

React allows us to write the standard html tags into a variable like this: `const AbilityTag = readonly ? "div" : "button"`, which then let's us render `<AbilityTag>` like any other component, representing either a `div` or a `button`. The uppercase name is important for JSX there.
This would then even let us get rid of the other changes the `readonly` prop tried to change to make the button not behave like a button, like the `.readonly` css class and `tabIndex={-1}`

ElementUser:
Thanks for all the feedback! Yes I agree accessibility is important

I'll have a look at this during the weekend when time permits

ElementUser:
@remmycat I've made the changes you requested (also, I did not know you can conditionally render HTML elements in that way in React, so that is cool - before I've always been re-defining the entire HTML element with its contents, so this is a nice little feature to mitigate code duplication).

Can you please have a look to see if everything is correct? I've done my own testing to make sure the behaviour still works as Sendou initially expects it to, and it does. I also tested tab indexing after the tabIndex content removal from the Ability icon to make sure it works as expected & it does on my end.

Thanks!

ElementUser:
Ah, silly me I forgot to re-push my changes to the correct branch. I have done so now

Sendouc:
This is looking nice! Just couple things we could think about still. The Abilities in AbilitySelector should probably still appear clickable as it's a functionality that you can click on them to remove the ability :)

I know this `readonly` was added before but now I'm actually thinking would it be a cleaner solution to just check for existance of `onClick` to decide whether it's readonly or not? I'm thinking this since it doesn't really make sense an ability has no `onClick` but is not readonly or vice versa.

ElementUser:
> This is looking nice! Just couple things we could think about still. The Abilities in AbilitySelector should probably still appear clickable as it's a functionality that you can click on them to remove the ability :)

Woops, you're absolutely correct. I missed that small behaviour since I don't use the build analyzer tool myself for Splatoon (not yet anyway :D). What I've done for the time being is conditionally set the class to readonly in `AbilitiesSelector` when it renders its `Ability` JSX elements:

`readonly={ability == "UNKNOWN" ? true : undefined}`

I don't think you want the unknown ability icons on the AbilitiesSelector to be clickable and/or draggable

ElementUser:
> I know this `readonly` was added before but now I'm actually thinking would it be a cleaner solution to just check for existance of `onClick` to decide whether it's readonly or not? I'm thinking this since it doesn't really make sense an ability has no `onClick` but is not readonly or vice versa.

I'll have a look at this

ElementUser:
Okay, I refactored the Ability component so that it renders the `readonly` class if no `onClick` prop is passed into it. 
- The `readonly` class is also rendered if the ability is an unknown one (since I presume there is no reason to be able to drag or click on an Unknown ability at this time).

@Sendouc whenever you get a chance, can you please have another look? Thanks!

ElementUser:
Hmm, looks like there is still an edge case if I try to do the `ability === "UNKNOWN"` logic - for some reason, the `AbilityTag` doesn't update to `button` or `div` when I drag an ability to the AbilitiesSelector :(. Will try to fix that edge case, or worst case scenario just allow UNKNOWN abilities to be treated as buttons for the time being

ElementUser:
Okay, I fixed it - I was just doing JavaScript's type check incorrectly for undefined. Should work in all cases now =)

Sendouc:
Yup this is nice 💯 I didn't consider about the UNKNOWN abilities before but it's very true they should not appear clickable also.

Issue #1068, opened 2022-10-29
If user has only 1 result, don't show the Set highlights section


Sendouc:
<img width="851" alt="image" src="https://user-images.githubusercontent.com/38327916/198830317-199904d0-10e7-487d-afd9-0cc88ab71bb9.png">

There is not really any benefit in doing so if you only have 1 result.

ElementUser:
Made a PR this one since it was quick: https://github.com/Sendouc/sendou.ink/pull/1071

PR #1071, opened 2022-10-29
Issue 1068:  Hide Highlights section if the user has exactly 1 result on their profile

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1068

# Description of Changes

Hide Highlights section if the user has exactly 1 result on their profile.

## Screenshots

See the comments for more detailed screenshots

### DIFF
--- a/app/routes/u.$identifier/results/index.tsx
+++ b/app/routes/u.$identifier/results/index.tsx
@@ -27,7 +27,8 @@ export default function UserResultsPage() {
   const isOwnResultsPage = user?.id === userPageData.id;
 
   const showHighlightsSection =
-    hasHighlights || (isOwnResultsPage && hasResults && userPageData.results.length !== 1);
+    hasHighlights ||
+    (isOwnResultsPage && hasResults && userPageData.results.length !== 1);
 
   return (
     <Main className="stack lg">
--- a/app/routes/u.$identifier/results/index.tsx
+++ b/app/routes/u.$identifier/results/index.tsx
@@ -27,7 +27,8 @@ export default function UserResultsPage() {
   const isOwnResultsPage = user?.id === userPageData.id;
 
   const showHighlightsSection =
-    hasHighlights || (isOwnResultsPage && hasResults);
+    (hasHighlights || (isOwnResultsPage && hasResults)) &&
+    userPageData.results.length !== 1; // Hide Highlights section if the user only has 1 result
 
   return (
     <Main className="stack lg">
--- a/app/routes/u.$identifier/results/index.tsx
+++ b/app/routes/u.$identifier/results/index.tsx
@@ -27,11 +27,7 @@ export default function UserResultsPage() {
   const isOwnResultsPage = user?.id === userPageData.id;
 
   const showHighlightsSection =
-    hasHighlights || (isOwnResultsPage && hasResults);
-
-  const showChooseHighlightsButton =
-    isOwnResultsPage &&
-    !(highlights.length === 1 && userPageData.results.length === 1);
+    hasHighlights || (isOwnResultsPage && hasResults && userPageData.results.length !== 1);
 
   return (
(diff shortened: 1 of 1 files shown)


app/routes/u.$identifier/results/index.tsx
```diff
@@ -27,7 +27,8 @@ export default function UserResultsPage() {
   const isOwnResultsPage = user?.id === userPageData.id;
 
   const showHighlightsSection =
-    hasHighlights || (isOwnResultsPage && hasResults);
+    (hasHighlights || (isOwnResultsPage && hasResults)) &&
+    userPageData.results.length !== 1; // Hide Highlights section if the user only has 1 result
```
remmycat, line 31: Thanks for looking at this!

I believe this logic might lead to issues with existing users who only have one result, which they've highlighted.
If these users exist, I believe the result would effectively be hidden until they get an additional result.

How about changing it, so that the highlight section is still always visible when `hasHighlights` is true? 🙂

ElementUser:
Oh I see, I thought you wanted the entire Highlights section gone & not just the `Choose Highlights` button. I'll make the adjustment now that you've clarified it for me

app/routes/u.$identifier/results/index.tsx
ElementUser: Oh I see, I thought you wanted the entire Highlights section gone & not just the `Choose Highlights` button. I'll make the adjustment now that you've clarified it for me :)

After adjusting it, here are the scenarios:

- 0 Highlights, 1 Result: the `Choose Highlights` button is displayed:
image

- 1 Highlight, 1 Result (though the Result is hidden once it's been highlighted, it is still in the results array): the `Choose Highlights` button is no longer there:

image

Any other scenario that doesn't match the other 2 options: the `Choose Highlights` button should be displayed (although I can't test this yet without figuring out how to add a new result to the database)

ElementUser: Okay, I managed to tinker with the DB & insert some data in the appropriate tables to get this test case:

- 1 Highlight, 2 total results (aka, 1 non-highlight): `Choose Highlights` button still shows:

image

remmycat: Oh, sorry, I think there was a misunderstanding.
If I understood @Sendouc correctly here, this issue is about: not showing the highlight section (including the choose highlights button) to a user if they have only one event result.

We basically want to disencourage people from marking a result as a "highlight", if it's their only one.

So your original solution was basically doing that.
My suggestion was to then still allow everyone to see your highlight (and edit it if it's your own), if you in the past did exactly the thing we're trying to prevent here.

So it's not just about the button, In my head I was thinking of a change from 
```typescript
const showHighlightSection = (hasHighlights || (isOwnResultsPage && hasResults)) && userPageData.results.length !== 1;
```
to:
```typescript
const showHighlightSection = hasHighlights || (isOwnResultsPage && hasResults && userPageData.results.length !== 1);
```

    
Note that we could additionally enforce all of these things on the backend side of things and also edit the DB to migrate everyone who has this scenario. But personally I don't see a need for that. If I got it correctly it's mostly about disencouragement of misusing this feature. @Sendouc?

ElementUser: Okay I see, my bad - yeah this one was harder for me to understand the intention of without additional info (and also I didn't initially see that the non-highlighted results does not equal the total results until much later on). Luckily, the fix is relatively simple & is just a one-liner like you mentioned :)

ElementUser: This is now what the page looks like if the user only has 1 total result:

image

If for some reason they had already highlighted their result, it will still render properly due to `hasHighlights` being true

remmycat (approved):
Nice, let's merge this! 💯

Issue #914, opened 2022-09-05
UserCombobox show just one result if exact discord id match


Sendouc:
Currently if you search user via Discord ID it just gives multiple:

<img width="404" alt="image" src="https://user-images.githubusercontent.com/38327916/196030511-c9440e21-03b8-40d1-8dac-c63452a1ba74.png">

ElementUser:
I'm reading some of the Fuse documentation & it seems that we can configure how "fuzzy" the logic is with several options.
- Reference: https://fusejs.io/api/options.html

Right now, I tried adjusting the `threshold` option and lowering it, and it seems to make things better overall. How low you want it to go totally depends on what you would like. I'll make a PR shortly so you can see what I did to make the search better in this specific case

PR #1075, opened 2022-10-30
UserComboBox is now more strict

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/914

# Description of Changes

Passed a FUSE_OPTIONS object to the Fuse() object.
- We can now adjust how "strict" the search is.
- We currently only adjust the `threshold` attribute of the FUSE_OPTIONS object

## Screenshot

With a `threshold` value of 0.42, this is what the search result looks like:

image

I tried with `threshold: 0.45` and it still gave more than 1 result, so I figured 0.42 is a good `threshold` value to start with.

## Reference

Docs for Fuse Options: https://fusejs.io/api/options.html

### DIFF
--- a/app/components/Combobox.tsx
+++ b/app/components/Combobox.tsx
@@ -68,12 +68,19 @@ export function Combobox<T extends Record<string, string | null | number>>({
     setLastSelectedOption(initialValue);
   }, [initialValue]);
 
+  // Reference for Fuse options: https://fusejs.io/api/options.html
+  const FUSE_OPTIONS = {
+    threshold: 0.42, // Empirically determined value to get an exact match for a Discord ID
+  };
+
   const filteredOptions = (() => {
     if (!query) return [];
 
     const fuse = new Fuse(options, {
+      ...FUSE_OPTIONS,
       keys: [...Object.keys(options[0] ?? {})],
     });
+
     return fuse
       .search(query)
       .slice(0, MAX_RESULTS_SHOWN)
--- a/app/components/Combobox.tsx
+++ b/app/components/Combobox.tsx
@@ -38,7 +38,7 @@ interface ComboboxProps<T> {
   clearsInputOnFocus?: boolean;
   onChange?: (selectedOption?: ComboboxOption<T>) => void;
   fullWidth?: boolean;
-  fuseOptions?: object;
+  fuseOptions?: Fuse.IFuseOptions<ComboboxOption<T>>;
 }
 
 export function Combobox<T extends Record<string, string | null | number>>({
@@ -53,7 +53,7 @@ export function Combobox<T extends Record<string, string | null | number>>({
   id,
   isLoading = false,
   fullWidth = false,
-  fuseOptions,
+  fuseOptions = {},
 }: ComboboxProps<T>) {
   const { t } = useTranslation();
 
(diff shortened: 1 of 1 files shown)


app/components/Combobox.tsx
```diff
@@ -68,12 +68,19 @@ export function Combobox<T extends Record<string, string | null | number>>({
...
+  // Reference for Fuse options: https://fusejs.io/api/options.html
+  const FUSE_OPTIONS = {
+    threshold: 0.42, // Empirically determined value to get an exact match for a Discord ID
+  };
+
   const filteredOptions = (() => {
     if (!query) return [];
 
     const fuse = new Fuse(options, {
+      ...FUSE_OPTIONS,
```
ElementUser, line 80: Currently, these settings apply for *all* comboboxes used in the application. I can do a refactor later on to optionally pass a FUSE_OPTIONS prop from the parent, this is more for proof-of-concept

ElementUser: Done the refactor

Sendouc:
Alright neat :D Didn't consider that but seems to work at least in this case. I'll leave a few small comments we should address.

app/components/Combobox.tsx
```diff
@@ -38,6 +38,7 @@ interface ComboboxProps<T> {
   clearsInputOnFocus?: boolean;
   onChange?: (selectedOption?: ComboboxOption<T>) => void;
   fullWidth?: boolean;
+  fuseOptions?: object;
```
Sendouc, line 41: There is pretty much never a good reason to use the `object` type in TS. When I hover over the "new Fuse" i can see what types it actually allows:

image

So this could be:

```suggestion
  fuseOptions?: Fuse.IFuseOptions<ComboboxOption<T>>;
```

app/components/Combobox.tsx
```diff
@@ -52,6 +53,7 @@ export function Combobox<T extends Record<string, string | null | number>>({
   id,
   isLoading = false,
   fullWidth = false,
+  fuseOptions,
```
Sendouc, line 56: Instead of doing

```ts
 if (typeof fuseOptions === "undefined") {
    fuseOptions = {};
  }
```

below it could be a bit cleaner to just simply do:

```suggestion
  fuseOptions = {},
```

Then if fuseOptions is nullish it gets empty object as default value.

app/components/Combobox.tsx
```diff
@@ -205,6 +213,11 @@ export function UserCombobox({
     );
   }
 
+  // Reference for Fuse options: https://fusejs.io/api/options.html
+  const USER_COMBOBOX_FUSE_OPTIONS = {
+    threshold: 0.42, // Empirically determined value to get an exact match for a Discord ID
+  };
```
Sendouc, line 219: Let's move this outside the function. Doesn't really matter but it's how I've been doing it and I find it cleaner as when you reading through the function it's not often relevant what exactly the `USER_COMBOBOX_FUSE_OPTIONS` is :)

Sendouc (changes requested):
Check comments but I still like this solution in itself 💯 Nice one

app/components/Combobox.tsx
ElementUser: I should definitely get used to highlighting to extract data types more in VSCode - thank you Sendou!

app/components/Combobox.tsx
ElementUser: Makes sense, I forgot about the default initializer for JS args - thanks again!

Sendouc:
💯

PR #1076, opened 2022-10-30
Update loc for DE

lolametro:


public/locales/de/analyzer.json
```diff
@@ -93,5 +100,15 @@
...
+  "torpedoExplanation": "Es ist unmöglich, mehr als einen Torpedo gleichzeitig abzufeuern (wegen der Spielmechanik).",
+  "trackingSubDefExplanation": "Die Markierungsdauern von Detektor, Tintenmine and Winkelmarker basieren auf einem Gegner mit 0 AP Sekundärstärke +.",
+  "distanceInline": "Reichweite: {{value}}",
+  "damageShort": "SCD",
+  "hitsToDestroyLong": "Treffer bis zerstört",
+  "hitsToDestroyShort": "TBZ",
+  "labels.amountOf": "Menge an",
+  "labels.damageType": "Schadenstyp",
+  "labels.weapon": "Waffe",
+  "dmgHtdExplanation": "SCD = Schaden • TBZ = Treffer bis zerstört",
```
lolametro, line 112: This formula `DMG = Damage • HTD = Hits to destroy` is a bit weird for me, at least mathematically speaking. How can damage be equal damage times HTD and that again be equal with HTD?

ElementUser: I agree here, I think `DMG` actually means "Damage needed to destroy".

So with that minor substitution, the formula would read as:

`Damage needed to destroy = Damage • HTD = Hits to destroy`

Slightly more concise:

`Damage to destroy = Damage • HTD = Hits to destroy`

lolametro: According to @Sendouc `it is just saying what the abbreviations mean, not a formula` 😅 The dot is just a separator.
Personally I'd either usenewline, or maybe semicolon ;
And/Or use colon: instead of equals sign =.

PR #1085, opened 2022-10-31
Only allow numbers as discrim

Illya9999:
Manually sending a request currently allows users to input anything that is 4 characters as their discrim which can be seen [here](https://sendou.ink/u/illya). I have added a simple check to make sure this is no longer possible.


Sendouc:
Nice catch :D

Issue #1032, opened 2022-10-21
Users search page


Sendouc:
Something like https://spl2.sendou.ink/u but exact implementation can be different.

Currently there is no way to search for a user on the site. Should be able to search by at least discord name, twitter name or splatoon 3 name (= in game name).

`useUsers` hook fetches all users and filtering can be done client side. Alternatively could do the filtering server side.

ElementUser:
I want to take a shot at this, though it will take a little longer than my usual PRs since this one looks more difficult to do thoroughly

ElementUser:
@Sendouc do you still have the code for that part of the web page? I should be able to adapt the dynamic list rendering part for the Splatoon 3 website. Tbh the fancy frontend part of the search is probably the hardest thing to get right :D

ElementUser:
Thank you @Sendouc 

I have a few questions after I started on it:

- Backend: does the `/users` endpoint get users with Twitter info or Splatoon 3 tag info? In my local client with the seeding, I don't seem to see that info anywhere: http://localhost:5800/users
- Frontend: do we use any CSS frameworks? From what I can see, we do not, and right now I'm starting off by porting the CSS manually for the UI component for the User Search Results

ElementUser:
@Sendouc also, here is the progress so far (only for the UI component):

![image](https://user-images.githubusercontent.com/15804376/199379064-4e8082ee-9fc0-4253-8c3a-6e0e98563591.png)

I spent quite a bit of time copying over the CSS & renaming everything so it makes sense (as well as renaming the CSS vars used by the old Splatoon 2 CSS)

Sendouc:
> * Backend: does the `/users` endpoint get users with Twitter info or Splatoon 3 tag info? In my local client with the seeding, I don't seem to see that info anywhere: http://localhost:5800/users

https://github.com/Sendouc/sendou.ink/blob/rewrite/app/routes/users.tsx

Logic is here. However adding fields can be problematic. Should check how big this is response is in prod and how much bigger adding new fields would make it approximately. But if not that then we should do the search serverside which of course is more complex and not fully sure how to approach that problem (although SQLite should be able to handle it)

> * Frontend: do we use any CSS frameworks? From what I can see, we do not, and right now I'm starting off by porting the CSS manually for the UI component for the User Search Results

We don't

ElementUser:
[...]

PR #1090, opened 2022-11-02
WIP:Users Search Page

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1032

# Description of Changes

Added Users Search Page (still a WIP btw)

## Other Changes

Added notes on fixing errors that the CI pipeline detects in `README.md`
- Also added a new npm command to fix CSS linting errors more easily: `npm run lint:styles-fix`

# WIP

Frontend
- Need to polish `UserSearchResult.tsx` UI component.
- Need to get the rendering React script written

Backend
- Need to ensure that I am pulling data from the correct place (regarding Twitter handles & what-not)


README.md
```diff
@@ -110,3 +110,21 @@ Note: This is only useful if you have access to a production running on Render.c
...
+npm run prettier:write
+```
+
+#### Stylelint step
+
+If you modified any `.css` or `.scss` files and the CI pipeline errors out on the `Stylelint` step, run this command in the repo's root directory:
+
+```sh
+npm run lint:styles-fix
+```
```
Sendouc, line 130: Should maybe instead point to `npm run cf` which runs all the same checks as CI and fixes them :) But it's true would be nice to document it.

app/components/UserSearchResult.tsx
```diff
@@ -0,0 +1,75 @@
+import { USER_SEARCH_PAGE } from "~/utils/urls";
+
+export function UserSearchResult({
```
Sendouc, line 3: This should be in the same file as route.

app/components/UserSearchResult.tsx
```diff
@@ -0,0 +1,75 @@
...
+              strokeWidth="0"
+              viewBox="0 0 512 512"
+              className="u__search_twitter-blue-icon"
+              height="1em"
+              width="1em"
+              xmlns="http://www.w3.org/2000/svg"
+            >
+              {/* Minified Twitter SVG - courtesy of: https://codepen.io/elliz/pen/BpqVzG */}
+              <path d="m492 110c-13 18-29 35-49 50c0 0 0 12 0 12c0 44-10 87-30 128c-21 41-53 76-96 104c-43 28-92 42-148 42c-55 0-104-14-149-43c5 1 13 1 24 1c45 0 85-13 120-40c-21-1-40-8-56-20c-17-12-28-28-34-48c3 1 9 2 17 2c9 0 18-1 26-3c-23-5-41-16-56-34c-14-18-21-38-21-61c0 0 0-1 0-1c12 6 27 11 43 12c-29-20-43-47-43-81c0-16 4-32 13-48c53 64 119 98 200 100c-2-6-3-13-3-21c0-27 9-50 28-68c19-19 42-28 69-28c28 0 51 9 70 29c20-4 41-11 61-22c-7 22-21 40-42 53c19-3 38-8 56-15" />
+            </svg>
```
Sendouc, line 63: components/icons folder for Icons

app/components/UserSearchResult.tsx
```diff
@@ -0,0 +1,75 @@
...
+export function UserSearchResult({
+  discordUsername,
+  discordId,
+  twitterHandle,
+}: {
+  discordUsername: string;
+  discordId: number;
+  twitterHandle?: string;
+}) {
+  const userPageHyperlink = `${USER_SEARCH_PAGE}/${discordId}`;
```
Sendouc, line 12: Check urls.ts

app/components/UserSearchResult.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  // Retrieves the string content before the '#' character in the full Discord username
+  let discordAvatarAltText: string | undefined;
+  const regexExtractDiscordNameWithoutIdentifier = new RegExp("^(.+?)#");
+  const regexArray =
+    regexExtractDiscordNameWithoutIdentifier.exec(discordUsername);
+  if (regexArray !== null) {
+    discordAvatarAltText = regexArray[0];
+  }
```
Sendouc, line 21: No need for alt text for the ava. The name itself as displayed on the screen describes it.

app/routes/u.tsx
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  return (
+    <Main className="stack lg">
+      <h1>{t("actions.search")}</h1>
+      <UserSearchResult
+        discordUsername={userData.discordUsername}
+        discordId={userData.discordId}
+        twitterHandle={userData.twitterHandle}
+      ></UserSearchResult>
```
Sendouc, line 30:
```suggestion
      <UserSearchResult
        discordUsername={userData.discordUsername}
        discordId={userData.discordId}
        twitterHandle={userData.twitterHandle}
      />
```

app/styles/u.css
```diff
@@ -169,3 +169,73 @@
...
+  -webkit-box-pack: center;
+  color: white;
+  font-size: calc(0.8rem);
+  font-weight: var(--semi-bold);
+  text-align: center;
+  text-transform: uppercase;
+  vertical-align: top;
+}
+
+.u__search_discord_avatar {
```
Sendouc, line 223:
```suggestion
.u-search__discord-avatar {
```

Should use this sort of naming convention instead to match other style files.

app/styles/vars.css
```diff
@@ -90,6 +90,7 @@ html.dark {
   --bg-mode-active: var(--theme-transparent);
   --abilities-button-bg: hsl(237.3deg 42.3% 26.6%);
   --border: hsl(237.3deg 42.3% 45.6%);
+  --border-radius-max: 9999px;
```
Sendouc, line 93: I don't think CSS var is needed for this

Sendouc (commented):
Left a few comments 💬

README.md
ElementUser: Oh, I didn't even know about that command - I will remove my command then & then update the README with that info. Thanks!

app/components/UserSearchResult.tsx
ElementUser: Done, and re-used the pre-existing Twitter icon too!

app/components/UserSearchResult.tsx
ElementUser: Hmm, VSCode is telling me that the `` tag needs an alt text of some kind - though I think that's from a plugin I use , jsx-a11y:

image

I'll still remove it though as you mentioned

app/routes/u.tsx
ElementUser: Surprised linter didn't catch this one - fixed

app/styles/vars.css
ElementUser: Fair enough, I was just porting CSS over from the Splatoon 2 user search page & this was one of the properties that was specified. I don't think it's needed either & will remove it

ElementUser: Okay, turns out that any kind of border-radius makes it round (so long as it is specified), so I went ahead with that in the `u.css` file instead

app/components/UserSearchResult.tsx
ElementUser:
```
C:\Users\ElementUser\_dev\Github\sendou.ink\app\routes\u.tsx
  61:11  warning  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text
```

IMO, we should include alt text after all, one of the linters is giving an error. And also I was mistaken, looks like jsx-a11y is actually something used by this project & isn't something different from my computer (the pipeline also outputs this warning)

ElementUser: For now, I changed the alt to just use the Discord Username (so no regex parsing or anything)

ElementUser:
@Sendouc : should we split up the issue into a few PRs? It might be easier that way.

Right now, I got the UI component mostly working. Future PRs could be with regards to tackling whether the data filtering should be done on server vs. client, and depending on the option chosen, whether we would need to do some code changes with regards to a User Search Combobox on its search algorithm (i.e., should we stick with fuseOptions & have it read from the users hook, or will it use a different hook to get filtered data from the serverside, etc.)?

As well as making a User Search Combobox variant to render this new UI component vs. what you have as a dropdown in the Admin panel

ElementUser:
Also: I found a bit of an issue with running npm run cf - it changed almost every single file cause of me being on Windows & the line-endings being different than the ones stored on Git (LF)

So.....maybe it's best not to ask users to run `npm run cf` on their end? And instead run the appropriate command on their end instead of doing everything at once

README.md
ElementUser: I found a bit of an issue with running npm run cf - it changed almost every single file cause of me being on Windows & the line-endings being different than the ones stored on Git (LF)

So.....maybe it's best not to ask users to run npm run cf on their end? And instead run the appropriate command on their end instead of doing everything at once

app/components/UserSearchResult.tsx
Sendouc: > or an empty string for decorative images

So in this case just doing alt="" would be correct I believe :) There is not really anything useful we can explain to screen reader users assuming we also have the user name next to it.

ElementUser: Fixed

ElementUser:
Sounds good - I'm going to put this on hold until the backend/data fetching & filtering is solidified. I can always adapt the components to whatever the data fetches afterwards, or if someone else is interested, they can work off my branch in the future. Either way works :)

README.md
ElementUser: Handled in these PRs:
- https://github.com/Sendouc/sendou.ink/pull/1092 
- https://github.com/Sendouc/sendou.ink/pull/1099

Sendouc:
Closing this for cleaning purposes.

Next step would be to implement the backend solution to better handle the current `<UserCombobox />` as described by Ilya here https://github.com/Sendouc/sendou.ink/issues/1032

Then we can revisit the actual user search page.

PR #1092, opened 2022-11-03
Add prettierrc and editorconfig

remmycat:


ElementUser:
Hey @remmycat , thanks for looking into this odd file line ending issue I was encountering with the default Git configs on Windows

I tested it on my end & from your branch, after running `npm run cf`  I see 63 files changed (instead of 400+). Those 63 files have CRLF endings before the change, and the line endings are changed to LF with the presence of your 2 config files you added to the repo.

If it's okay with you, I can make a commit for these changed files (they only have changed line endings & they are set to LF) onto your branch (or a fork of it, if I can't be given some kind of contributor permissions to this project).

Thanks again!

remmycat:
Huh, I see! Thanks for testing this!
Yes, please add the commit to the branch and I'll have a look :) 
(You're showing up as Contributer to me here, so you should be allowed now?)

remmycat:
Ah, sorry, my mistake - Got caught up on the wording. Contributor only means your email address is in the commit history, "Collaborator" comes with permissions.

Feel free to ask Sendou ofc. But I don't mind if it's a branch in your repo instead.

remmycat:
@ElementUser Sorry to bug you again, but do you think you could try checking out the newest state of this branch with your gitconfig reverted? 😅  The documentation you linked and the rest of the internet say that the .gitattributes file with `* text=auto` should fix this issue (effectively overwriting the user's auto crlf setting), when checking out the repo.

I added a bunch of test changes, so there are some modified files. 

If things are still behaving weird, I don't see any issues with just merging this (without the last commit) and leaving the explanation in the Readme. But I'd prefer the issue not coming up again at all 🙂

ElementUser:
image

Sadly, the same thing happens after I switched back to this repo's URL as the origin, then ran `git fetch` and `git pull`, fixed any merge errors, committed, then ran `npm run cf` again & got 50 files that appear to have changed =(

Once again, running `git add --renormalize .` fixes it for me. I don't think it's harmful to commit what you have (minus those random text changes you put in)

remmycat:
Alright then! Thanks a lot for trying it again

PR #1097, opened 2022-11-05
Wrap useTranslation to detect missing translations for the route

### LINKED ISSUE #1080: More safe useTranslation hook
- Wrap useTranslation
- In dev check handle that correct translations are loaded and throw error if not (there is Remix hook for checking handle data I believe and from useTranslation input we know what it expects)

remmycat:
Closes #1080


.eslintrc.js
```diff
@@ -31,6 +31,19 @@ module.exports = {
...
+        paths: [
+          {
+            name: "react-i18next",
+            importNames: ["useTranslation"],
+            message:
+              "Please import useTranslation from '~/hooks/useTranslation' instead.",
+          },
+        ],
+      },
+    ],
```
Sendouc, line 46: Cool way to handle it. Maybe later we could also use this to prevent importing from deep in the modules folder 🤔 so only allow modules/xxx imports

Sendouc (approved):
This is so nice! I have this bug all the time so will be much easier to detect it. + of course for new contributors will be so useful.

.eslintrc.js
remmycat: Yes! I only recently learned about it and it's quite helpful for situations like these.

One really nice usecase I sometimes have is also if you have packages installed only for importing types from them, but it would be a big problem if the actual import landed in the bundle somehow (e.g. because of size, or because it has node-only code, or it would create a bad circular dependency or something).

With `@typescript/eslint/no-restricted-imports` you can then forbid importing from that package, but [ set `allowTypeImports: true`](https://typescript-eslint.io/rules/no-restricted-imports/#allowtypeimports), and it won't complain for any `import type`.

PR #1098, opened 2022-11-05
Make LanguageChanger respect existing search params

### LINKED ISSUE #1096: Switching the language on the map list page resets map pool parameters
Found by @DoubleCookies:
>Another thing I found:
In map lists changing language cause removing pool parameters from URL. It can be restored by clicking on any mode icon twice, but still probably parameters should append, not overwrite each other
>
> [sendou.ink/maps?pool=sz%3A1bea%3Btc%3A1f0f%3Brm%3A1f86%3Bcb%3A1c3d]()
> **\*click\*** Language -> Français
> [sendou.ink/maps?Ing=fr]()

remmycat:
Closes #1096 

This fixes the issue, but it feels a little bit brute-force. I'm not sure what a better solution would be 🤔 

I guess one option would be not to use links at all, but to persist the language using a fetch/form and then somehow invalidate the root loader cache? (possibly by still setting the search param programatically, or maybe there is another option)

### DIFF
--- a/app/components/layout/LanguageChanger.tsx
+++ b/app/components/layout/LanguageChanger.tsx
@@ -1,11 +1,24 @@
+import { useSearchParams } from "@remix-run/react";
 import { useTranslation } from "~/hooks/useTranslation";
 import { languages } from "~/modules/i18n";
 import { LinkButton } from "../Button";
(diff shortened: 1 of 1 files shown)


Sendouc (approved):
I don't see a problem with this solution :)

PR #1101, opened 2022-11-05
Improve plus suggestion comment validation

### LINKED ISSUE #828: Plus suggestion body not trimmed before saving to database
image

Should not be possible to have white space at the end or start: " " or "\n" or whatever. Solution -> call trim on the server before saving it to datatabase.

Compare to one without newline at the end and you see the difference:

image

remmycat:
Closes #828 
Just learned about this "native" validation API, I love it :)


Sendouc:
💯 💯 💯 

Very nice! I knew about the custom validations but thought they were hard/unwieldy to use for some reason. You are showing otherwise :D

Sendouc (approved):

PR #1102, opened 2022-11-06
Add placement ordinal translation

### LINKED ISSUE #1016: Translate/remove ordinals for non-english languages
See discussion in https://github.com/Sendouc/sendou.ink/pull/1015

Either translate them for the languages we support or remove them which is totally fine as well.

remmycat:
Closes #1016 

I hope the creative use of `^` to detect superscript is fine. Pretty sure there are no languages that show "1^ place" 😅 

I tried adding some translations for languages where I could find a clear answer.

References: https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html + google results on "\<language\> rank ordinal suffix" or similar.
Unfortunately the unicode comparison alone doesn't go into detail in regards to grammatical gender. In some cases the grammatical gender of the described object changes the suffix, but in most cases this should then get the same gender as the translation for "place".

### DIFF
--- a/scripts/check-translation-jsons.ts
+++ b/scripts/check-translation-jsons.ts
@@ -4,6 +4,8 @@ import prettier from "prettier";
 
 const NO_WRITE_KEY = "--no-write";
 
+const KNOWN_SUFFIXES = ["_zero", "_one", "_two", "_few", "_many", "_other"];
+
 const dontWrite = process.argv.includes(NO_WRITE_KEY);
 
 const otherLanguageTranslationPath = (code?: string, fileName?: string) =>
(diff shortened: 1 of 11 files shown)


scripts/check-translation-jsons.ts
```diff
@@ -188,6 +197,48 @@ function validateNoDuplicateKeys({
   }
 }
 
+// get keys while respecting different plural/context key suffixes in different languages.
+function getKeysWithoutSuffix(
+  translations: Record<string, string>,
+  lang: string,
+  file: string
+): string[] {
+  // Map<baseKey, Set<suffixes>>
```
remmycat: Ah, yes, it's a leftover! I had a Map there before and tried to explain the key and value. Will remove.

Sendouc, line 206: Is this comment leftover or maybe I don't understand it 🤔

Sendouc (approved):
Nice!

Issue #1104, opened 2022-11-06
Translation Progress


remmycat:
This is supposed to become an automatically edited issue keeping track of translation progress, just like https://github.com/remmycat/sendou.ink-contrib/issues/1

(PR will follow shortly)

PR #1105, opened 2022-11-06
Automatically track translation progress inside issue

remmycat:
This change should get rid of the annoying task of always keeping the translation progress up to date.

The `translation-progress.md` will no longer be checked in, but the PR pipeline still runs the `check-translation-jsons:no-write` script to make sure we don't have any issues in the translation files. (The script no longer errors if the existing file doesn't exist / is different)

Instead, a new GitHub Action will update this issue: #1104 , which we can pin and possibly lock

I also tried to create a better overview table 🙂  ~~and used a little React/HTML where the markdown would've gotten silly otherwise~~ (reverted)

See https://github.com/remmycat/sendou.ink-contrib/issues/1 for an example of the output


remmycat:
One thing I want to investigate before merging this is if it's really necessary to go for an html table here.

In previous tries I had put the `<details>` elements directly into the table cells, which turned out to be an awful idea. But there markdown tables were definitely not enough.

Now that it's really just the emoji and text in each cell, I think getting the HTML (and usage of React for rendering) out of this again might be much neater 🤔

remmycat:
Done, back to markdown everything, output is much nicer to look at again.

Sendouc (approved):

Issue #1115, opened 2022-11-09
Make map pool section text able to be translated


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/200874536-b74b7d60-1d2a-416f-a3eb-e42fccb4bc0f.png)

PR #1118, opened 2022-11-11
Fixed "Map Pool" title in events to use the appropriate localization value

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1115

# Description

Fixed "Map Pool" title on active event pages to use the appropriate localization value

## Screenshot of the bug before the fix is applied:

image

## Screenshot of the bug fix on an active event page:

image


Sendouc:
👍

Issue #1117, opened 2022-11-10
Calendar week links widths not symmetric


Sendouc:
<img width="469" alt="image" src="https://user-images.githubusercontent.com/38327916/201196100-2e67ba6e-bafa-4ee8-ba29-60a101b00fb4.png">

The first one from left and last are different width. Should be symmetrical so they are same width and then second from left and second last are same width etc.

ElementUser:
Ah dang, didn't realize my previous fix didn't end up fixing this issue - I might find time to take a look at this again soon (hopefully this week)

Sendouc:
I will probably just end up reworking the whole page to something a bit different down the line.

Issue #935, opened 2022-09-08
French localization WeekLinks out of space


Sendouc:
Shorter translation? Adjust CSS?

![Image](https://user-images.githubusercontent.com/38327916/189164561-dc94b3e7-32b3-4b73-a8ed-6aa63f139a83.png)

DoubleCookies:
Btw not same, but a bit annoying word break in ru locale
![image](https://user-images.githubusercontent.com/12849336/196382455-21394efc-2a6f-4bf3-bcda-e59d6e61e9dc.png)

ElementUser:
Adjusted CSS to resolve this issue as well in this PR: https://github.com/Sendouc/sendou.ink/pull/1119

ElementUser:
@DoubleCookies I can't read Russian, but I believe my changes in PR !1119 also addresses the edge case for RU strings:

![image](https://user-images.githubusercontent.com/15804376/201445990-7fc2337f-8702-4960-b027-f900b932776d.png)

I will include this image in my PR as well.

PR #1119, opened 2022-11-11
Fixed a few calendar WeekLink element issues

ElementUser:
# Linked Issues

https://github.com/Sendouc/sendou.ink/issues/935
https://github.com/Sendouc/sendou.ink/issues/1117

# Description of  Changes

Fixed calendar week link element widths not being symmetric for the left-most and right-most elements (https://github.com/Sendouc/sendou.ink/issues/1117)

I had to slightly increase the size of the 3rd and 7th (left-most and right-most visible) calendar link elements to fix this as properly as possible to accommodate different languages & screen resolutions. Keeping the dimensions the same as the live website for the smaller calendar link element simply wasn't feasible.

Additionally, I have to overwrite `min-width` and `max-width` to still properly keep the text inside the calendar link element for certain languages (e.g., Japanese). If I only defined a `width` element for the 3rd and 7th child elements (and did not redefine `min-width` and `max-width`), we would get a visual bug on certain screen resolutions for certain languages. Example:

image

I also fixed the French localization (and most likely fixed the Russian localization) edge case(s) for string contents overflowing the element container by slightly increasing the max-width (for: https://github.com/Sendouc/sendou.ink/issues/935). This still looks fine on 375px width screens.

## Testing Notes

I briefly tested this fix for all languages & at a reasonably-sized phone width in the web browser dev tools tab (at a width of 375px), and everything lines up well. I also briefly scrolled left and right a reasonably sane amount to check that there weren't any obvious edge cases.

## Screenshots

- Smallest phone screen width on Google Chrome dev tools, English language:
image

- Smallest phone screen width on Google Chrome dev tools, Japanese language:
image

- Smallest phone screen width on Google Chrome dev tools, French language (for https://github.com/Sendouc/sendou.ink/issues/935):
image

- Smallest phone screen width on Google Chrome dev tools, Russian language (for https://github.com/Sendouc/sendou.ink/issues/935):
image

- Regular screen, English:
image

- Regular screen, Japanese:
image

### DIFF
--- a/app/styles/calendar.css
+++ b/app/styles/calendar.css
@@ -7,6 +7,13 @@
   margin-inline-start: auto;
 }
 
+.calendar__week__relative {
+  max-width: 3rem;
+  margin: 0 auto;
+  text-align: center;
+  white-space: pre-line;
+}
+
 .calendar__weeks {
   display: flex;
   align-items: center;
@@ -103,13 +110,6 @@
 .calendar__week:nth-child(7) {
   min-width: calc(var(--full-size-week-width) - 1.5rem);
   max-width: calc(var(--full-size-week-width) - 1.5rem);
-}
-
-.calendar__week__relative {
-  max-width: 3rem;
-  margin: 0 auto;
-  text-align: center;
-  white-space: pre-line;
 }
 
 .calendar__week__dash {
--- a/app/styles/calendar.css
+++ b/app/styles/calendar.css
@@ -1,4 +1,3 @@
-/* stylelint-disable no-descending-specificity */
 .calendar__placeholder {
   height: 48rem;
 }
@@ -73,7 +72,7 @@
 /* Addresses spacing issues for translations with slightly longer strings
  * Issue: https://github.com/Sendouc/sendou.ink/issues/935
  */
-.calendar__week:nth-child(5) > .calendar__week__relative {
+.calendar__week:nth-child(5) .calendar__week__relative {
   max-width: 3.6rem;
 }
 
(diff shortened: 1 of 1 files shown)


app/styles/calendar.css
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable no-descending-specificity */
```
Sendouc, line 1: I don't think we should disable the linter here unless I'm missing something 🤔

ElementUser: Thanks for the feedback, I had a second look & looked up a CSS Selectors reference. Turns out all I needed to do was remove the `>` symbol and my fix still functions as intended. There is a very subtle difference between `.classname > .classname` and `.classname .classname`

ElementUser: @Sendouc nevermind, linter still fails :(

https://github.com/Sendouc/sendou.ink/actions/runs/3451373555/jobs/5760549697

I actually do not know how to fix this unless we disable the linter rule. It's an odd CSS Selector issue where if I rearrange the order like the linter asks me to, my fix ends up not working at all & not fixing the originally reported issues.

ElementUser: @Sendouc I had another look today & found out I didn't have to disable the rule after all!

All I have to do is make sure that the CSS class with lower specificity (`.calendar__week__relative`) is defined earlier in the .css file than everywhere else that it is used as a higher specificity (i.e., deeper level) CSS property.

Reference: https://stackoverflow.com/a/57871548

So it all works out now & it should be good to be merged whenever you're ready. I double-checked that my fix still works & indeed it does!

Issue #1093, opened 2022-11-04
Show how many chunks to make build


Sendouc:
<img width="1118" alt="image" src="https://user-images.githubusercontent.com/38327916/200047570-196c94a5-9e54-44cc-9577-200c42095b38.png">

ElementUser:
This didn't have a `help wanted` tag on it, but this one didn't look too difficult to do, so I went ahead & implemented the requested feature in this PR: https://github.com/Sendouc/sendou.ink/pull/1120

PR #1120, opened 2022-11-11
Added Ability Chunks Required section in Build Analyzer

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1093

# Description of Changes

Added an `Ability Chunks Required` section below the `Ability Points` section on the Build Analyzer page.

This section correctly calculates the amount of ability chunks required for the current build if at least 1 ability is selected in the build. This includes handling primary slot-only abilities like Ninja Squid, Tenacity, Object Shredder, etc.

We build a map data structure of `<Ability Chunk Name, Number of Chunks Required>`, iterate through each piece of gear on the build and handle the main and sub abilities separately, while filtering out "UNKNOWN" entries to ignore them.
- To handle the primary slot-only abilities like Ninja Squid, I added a new attribute to each object in `abilities` called `abilityChunkTypesRequired`, which is an array of ability names.

With specific ability chunk counts for abilities (45 for a stackable ability, 15 for a primary-slot-only required ability), I then iterated through each main ability. If it's a stackable ability, then add its key to the map & add 45 to the pre-existing value. If it's a primary-slot-only ability, extract the array of ability names from the `abilityChunkTypesRequired` property of the main ability and iterate through those, add its key to the map & add 15 to the pre-existing value.

For stackable sub abilities, we add 10 to the pre-existing value instead. The ability chunk rule for the same ability on the same piece of gear is followed (1 ability = 10 chunks, 2 abilities = 20 chunks, 3 abilities = 30 chunks).
- A temporary ability chunk map is used to keep track of this progress for each individual piece of gear as the code iterates through each piece of gear. 

Before rendering the visual elements pertaining to this map, the map data structure is converted to an array & then sorted by their value (descending), followed by their abbreviation names (in ascending order). Value is the number of ability chunks required.

## Screenshots of Examples

3 Main Stackable abilities:

image

2 Stackable abilities, 1 Primary Slot-Only ability:

image

3 Primary Slot-Only abilities:

image

An actual full build that I would use as a Slayer (that also conveniently tests most scenarios, with each piece of gear having either 1, 2 or 3 of the same types of sub abilities):

image

## Unit Tests

I've added unit tests in `app\modules\analyzer\abilityChunksCalc.test.ts`. I also added a helper function to check if the final value contains all of the entries that we expect it to in an order-agnostic way (since we do not want to worry about ordering/sorting in these tests).
- A detailed error printout assists with determining the cause of the failing test.

# References for Ability Chunks
- https://splatoonwiki.org/wiki/Ability_chunk#Splatoon_3


ElementUser:
1 question I wanted to clarify:

- Is `BRU` Sub Power Up? The reason I ask is because `SPU` can stand for `Special Power Up` as well as `Sub Power Up`, and wanted to make sure I got this correct.

app/routes/analyzer.tsx
```diff
@@ -735,6 +746,83 @@ function AbilityPointsDetails({
...
+    const mainAbility = mainAbilities[i];
+    if (typeof mainAbility !== "undefined" && mainAbility !== "UNKNOWN") {
+      const primarySlotOnlyAbilityRef = abilities.filter(
+        (a) => a.name === mainAbility && a.abilityChunkTypesRequired.length > 0
+      );
+
+      // Primary-slot-only item that can have ability chunks put on it
+      if (primarySlotOnlyAbilityRef.length === 1) {
+        const primaryAbility = primarySlotOnlyAbilityRef[0];
+        if (typeof primaryAbility !== "undefined") {
```
ElementUser, line 768: I end up having to use a lot of these `if (typeof variableName !== "undefined")` checks. Is there a better way of doing this in my current use cases? TypeScript forces me to add these checks for type safety, and I'm just wondering if there's a less cluttered way of doing these checks.

Sendouc:
> Is BRU Sub Power Up? The reason I ask is because SPU can stand for Special Power Up as well as Sub Power Up, and wanted to make sure I got this correct.

BRU = Sub Power Up ("Bomb Range Up")
SPU = Special Power Up

app/routes/analyzer.tsx
```diff
@@ -735,6 +751,88 @@ function AbilityPointsDetails({
...
 }
 
+/**
+ * From an array of Main abilities, create a map of <Ability, number>, then return it as an Array after sorting by value, descending.
+ *
+ * The map describes the number of Ability chunks required to replace the main ability on a piece of gear.
+ *
+ * Extra processing is required for Main abilities that are primary slot-only abilities,
+ *  as they are comprised of 3 stackable ability chunks at a lower ability chunk count than usual.
+ */
```
Sendouc, line 761: This shouldn't be a JSDOC comment since those should be used for describing to the caller what the function is. Regular comment looks good in this case.

app/routes/analyzer.tsx
```diff
@@ -735,6 +751,88 @@ function AbilityPointsDetails({
...
+function getAbilityChunksMapAsArray(mainAbilities: AbilityWithUnknown[]) {
```
Sendouc, line 762: Let's move this function to the analyzer module (new file) + maybe add test or two for it.

app/routes/analyzer.tsx
```diff
@@ -735,6 +751,88 @@ function AbilityPointsDetails({
...
+        abilityChunksMap.set(
+          mainAbility,
+          (abilityChunksMap.get(mainAbility) ?? 0) +
+            REQUIRED_ABILITY_CHUNKS_COUNT
+        );
+      }
+    }
+  }
+
+  // After building the map, convert it to an array & sort the map by value, descending
```
Sendouc, line 801:
```suggestion
```

Code explains itself quite well.

app/routes/analyzer.tsx
Sendouc: You are inside a loop. So I think it would make it quite a bit cleaner to just do: 

```ts
if (!variableName) continue

// logic continues here
```

app/routes/analyzer.tsx
```diff
@@ -735,6 +751,88 @@ function AbilityPointsDetails({
...
+      );
+
+      // Primary-slot-only item that can have ability chunks put on it
+      if (primarySlotOnlyAbilityRef.length === 1) {
+        const primaryAbility = primarySlotOnlyAbilityRef[0];
+        if (typeof primaryAbility !== "undefined") {
+          for (
+            let j = 0;
+            j < primaryAbility?.abilityChunkTypesRequired.length;
+            j++
```
Sendouc, line 778: Do this instead:

```suggestion
          for (const ability of primaryAbility.abilityChunkTypesRequired)
```

Also: if you have write `j < primaryAbility?.abilityChunkTypesRequired.length` something went wrong since you shouldn't be comparing number and `undefined`

Sendouc (changes requested):
Hey really nice you took a shot at this :) I think it's looking quite good already but check the comments I left on the code. For the implementation of the feature I think we should calculate chunks required to make the whole build, not just main abilities.

ElementUser:
> Hey really nice you took a shot at this :) I think it's looking quite good already but check the comments I left on the code. For the implementation of the feature I think we should calculate chunks required to make the whole build, not just main abilities.

Surprisingly, I did not know you can actually change the sub abilities :D I'm not sure why I never tried lol, I think it's because I was always interested in replacing the main ability of my gear. Thanks for the feedback, code was written in a way where I can easily include the sub-abilities - I just need a way to differentiate between main ability & sub ability in the calculations is all

app/modules/in-game-lists/abilities.ts
```diff
@@ -1,30 +1,81 @@
...
-  { name: "CB", type: "HEAD_MAIN_ONLY" },
-  { name: "NS", type: "CLOTHES_MAIN_ONLY" },
-  { name: "H", type: "CLOTHES_MAIN_ONLY" },
-  { name: "TI", type: "CLOTHES_MAIN_ONLY" },
-  { name: "RP", type: "CLOTHES_MAIN_ONLY" },
-  { name: "AD", type: "CLOTHES_MAIN_ONLY" },
-  { name: "SJ", type: "SHOES_MAIN_ONLY" },
-  { name: "OS", type: "SHOES_MAIN_ONLY" },
-  { name: "DR", type: "SHOES_MAIN_ONLY" },
+  { name: "ISM", type: "STACKABLE", abilityChunkTypesRequired: [] },
```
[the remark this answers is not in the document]
ElementUser: We can do this, the only thing that I'll require in the code is a check to see if that property exists when attempting to use it:

```js
const primarySlotOnlyAbilityRef = abilities.filter(
        (a) => {
          if (Object.prototype.hasOwnProperty.call(a, 'abilityChunkTypesRequired')) {
            return a.name === mainAbility && (a.abilityChunkTypesRequired.length > 0)
          }
          return false;
        }
      );
```
Without this check, we get a `TypeError: Cannot read properties of undefined (reading 'length')` error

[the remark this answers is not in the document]
ElementUser: An alternative I could do to prevent us having to do a key/property check is create a separate data structure that only contains a `mainAbilityName: subAbilityNames[]` map data structure and then do some data structure linkages, but I feel that keeping everything in the `abilities` const is cleaner at the moment

app/routes/analyzer.tsx
ElementUser: Great idea, I forgot `.ts` is supposed to contain more detailed/in-depth logic & `.tsx` is supposed to contain DOM-specific logic for the most part.

I did this & am in the process of adding tests

ElementUser: @Sendouc I added some meaningful tests just now - can you please take a look when you have time? Thanks =)

ElementUser:
@Sendouc : I updated the code to process both main abilities and sub abilities properly now. It looks very nice & I updated the screenshots in the main post, as well as added some unit tests for these test cases =)

ElementUser:
Oh no, I just realized that there is more nuance to ability chunk upgrading......bummer :(. Goes to show how often I do stuff with Ability Chunks...

So for a piece of gear, assuming we have 1 sub of a specific type on it already (let's say Swim Speed Up since that's popular), to add another SSU sub, we need 20 ability chunks (not 10). And if we already have 2 subs of a specific type on the gear & we want to add a third one, then we need 30 ability chunks.

As there is no way of us knowing which pieces of gear already has which amount of subs, should we add some kind of a disclaimer saying that the total ability chunk count shown is under these base assumptions? I kind of understand why Ability Chunks required for sub abilities is a lot less meaningful now & it's another reason why Main Ability chunks are a little more accurate.

@Sendouc any advice on what to do next? We could also ask the original issue reporter to get his opinion on how he envisions this feature to work.

Sendouc:
> Oh no, I just realized that there is more nuance to ability chunk upgrading......bummer :(. Goes to show how often I do stuff with Ability Chunks...
> 
> So for a piece of gear, assuming we have 1 sub of a specific type on it already (let's say Swim Speed Up since that's popular), to add another SSU sub, we need 20 ability chunks (not 10). And if we already have 2 subs of a specific type on the gear & we want to add a third one, then we need 30 ability chunks.
> 
> As there is no way of us knowing which pieces of gear already has which amount of subs, should we add some kind of a disclaimer saying that the total ability chunk count shown is under these base assumptions? I kind of understand why Ability Chunks required for sub abilities is a lot less meaningful now & it's another reason why Main Ability chunks are a little more accurate.
> 
> @Sendouc any advice on what to do next? We could also ask the original issue reporter to get his opinion on how he envisions this feature to work.

🤔 I don't see it as a problem. You just calculate it assuming the unknown slots are empty.

So if row has 

Swim Speed / ? / ? then needs 10 swim speed chunks

if

Swim Speed / ? / Swim Speed then needs 10+20 = 30 swim speed chunks.

If people want to only consider main then they will only fill main slots.

ElementUser:
> > Oh no, I just realized that there is more nuance to ability chunk upgrading......bummer :(. Goes to show how often I do stuff with Ability Chunks...
> > So for a piece of gear, assuming we have 1 sub of a specific type on it already (let's say Swim Speed Up since that's popular), to add another SSU sub, we need 20 ability chunks (not 10). And if we already have 2 subs of a specific type on the gear & we want to add a third one, then we need 30 ability chunks.
> > As there is no way of us knowing which pieces of gear already has which amount of subs, should we add some kind of a disclaimer saying that the total ability chunk count shown is under these base assumptions? I kind of understand why Ability Chunks required for sub abilities is a lot less meaningful now & it's another reason why Main Ability chunks are a little more accurate.
> > @Sendouc any advice on what to do next? We could also ask the original issue reporter to get his opinion on how he envisions this feature to work.
> 
> 🤔 I don't see it as a problem. You just calculate it assuming the unknown slots are empty.
> 
> So if row has
> 
> Swim Speed / ? / ? then needs 10 swim speed chunks
> 
> if
> 
> Swim Speed / ? / Swim Speed then needs 10+20 = 30 swim speed chunks.
> 
> If people want to only consider main then they will only fill main slots.

@Sendouc done! Updated the code to do the correct calculations, updated the main post, and also updated some tests. I added 2 "real builds" (as in, builds I would personally use) for good measure.

Whenever you have time, please review again & let me know if it's good to go =)

app/routes/analyzer.tsx
ElementUser: Updated tests again btw (they are really good now)

Sendouc (commented):
This looks really nice now! 💯 great job with the tests!

One final thing before merging: the accordion should be hidden before we select first ability (matching behavior with other accordion).

ElementUser:
> This looks really nice now! 💯 great job with the tests!
> 
> One final thing before merging: the accordion should be hidden before we select first ability (matching behavior with other accordion).

Hi sendou, thanks for the reply!

I found an edge case & if we use the same condition to show/hide the accordion element for ability chunks & I select main abilities only, then it won't show the ability chunks section.

What I can do is use a filter() function to check if the builds array of arrays contains an element other than "UNKNOWN" (similar to what I wrote in the test file). I'll do that later on today

ElementUser:
@Sendouc issue should be fixed now - it will now behave as we expect & also fixes the edge case I brought up with primary slot-only abilities like Ninja Squid.

Test cases:

- 0 abilities chosen (Ability Chunks accordion is hidden):

image

- 1 ability is chosen (any slot):

image

- 1 primary slot-only ability is chosen (in a main ability slot):

image

Sendouc:
💯

Issue #1053, opened 2022-10-27
Flicker with combobox


Sendouc:
https://user-images.githubusercontent.com/38327916/198339006-94f7177e-6b22-4a45-bbf3-33299538923c.mov

ElementUser:
I looked at replicating the issue & found out that it only happens the first time you search for something & do a selection after you've loaded onto that page. If you repeat this without refreshing or leaving the page, this bug won't happen.

ElementUser:
In `Combobox.tsx`, if I set `initialValue={undefined}`, I cannot replicate the issue anymore

ElementUser:
Hmm, nvm on some of that - my PC performance is too good, so that explains why I seem inconsistent on seeing this issue.

It is more prominent on my phone (since it has less computing power than my main pc)

ElementUser:
Okay, I found an actual solution (though probably not the one you want): forcibly preventing reloads in `analyzer.tsx` actually works! I can no longer replicate the original bug if I add this line to the file:

`export const unstable_shouldReload = () => false;`

Reference: https://remix.run/docs/en/v1/api/conventions#never-reloading-the-root

I am sure there is a better way to resolve this though, but leaving this here for your info

ElementUser:
@Sendouc I got a viable solution in this PR: https://github.com/Sendouc/sendou.ink/pull/1131

I explain everything in detail there - hopefully that helps!

PR #1131, opened 2022-11-17
Fixed flicker with combobox issue using unstable_shouldReload()

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1053

# Description

Fixed a flicker issue with the Weapon Combobox component.

To fix this, I've added `unstable_shouldReload()` in `app\routes\analyzer.tsx`.

# Manual testing

To replicate the bug reliably in the live version of the website, I enabled low-end mobile throttling in my Google Chrome's web browser debugging tools. By doing this, I can visibly see the flicker every time I change my search selection in the Weapon Combobox.

I did the same thing after implementing my fix on the locally hosted application & could no longer replicate this issue.

# References
- Remix docs with regards to the `unstable_shouldReload()` function https://remix.run/docs/en/v1/api/conventions#never-reloading-the-root
- Discussion on the Remix Github project that talks about this issue: https://github.com/remix-run/remix/discussions/2700

### DIFF
--- a/app/routes/analyzer.tsx
+++ b/app/routes/analyzer.tsx
@@ -67,9 +67,7 @@ export const handle: SendouRouteHandle = {
 };
 
 // Resolves this Github issue: https://github.com/Sendouc/sendou.ink/issues/1053
-export const unstable_shouldReload: ShouldReloadFunction = ({ url }) => {
-  return Boolean(url) && !url.searchParams.get("weapon");
-};
+export const unstable_shouldReload: ShouldReloadFunction = () => false;
 
 export default function BuildAnalyzerPage() {
   const { t } = useTranslation(["analyzer", "common", "weapons"]);
--- a/app/routes/analyzer.tsx
+++ b/app/routes/analyzer.tsx
@@ -1,4 +1,5 @@
 import { type LinksFunction, type MetaFunction } from "@remix-run/node";
+import type { ShouldReloadFunction } from "@remix-run/react";
 import { Link } from "@remix-run/react";
 import * as React from "react";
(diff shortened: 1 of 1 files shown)


app/routes/analyzer.tsx
```diff
@@ -65,6 +66,11 @@ export const handle: SendouRouteHandle = {
   navItemName: "analyzer",
 };
 
+// Resolves this Github issue: https://github.com/Sendouc/sendou.ink/issues/1053
+export const unstable_shouldReload: ShouldReloadFunction = ({ url }) => {
+  return Boolean(url) && !url.searchParams.get("weapon");
+};
```
Sendouc, line 72: Good findings!! Should we just do:

```suggestion
export const unstable_shouldReload: ShouldReloadFunction = () => false;
```

since analyzer doesn't have a `loader` function set I believe this would be correct :) Although idk why Remix doesn't do this for us.

ElementUser: Sure thing, let's do that instead then - I wasn't sure if we wanted to hard set it to `false` initially because I wasn't 100% on whether this function affects anything other than a loader, so thank you for checking.

> Although idk why Remix doesn't do this for us.

In the docs, they mention that they default the behaviour to reload everything to be safe: https://remix.run/docs/en/v1/api/conventions#unstable_shouldreload

> During client-side transitions, Remix will optimize reloading of routes that are already rendering, like not reloading layout routes that aren't changing. In other cases, like form submissions or search param changes, Remix doesn't know which routes need to be reloaded so it reloads them all to be safe. This ensures data mutations from the submission or changes in the search params are reflected across the entire page.

PR #1133, opened 2022-11-18
update translation (JA)

aisuzuki:
Three new items are added into translation files.
I wonder if it would be better for you to wait a bit until number of untranslated items becomes certain amount in order to avoid constant server deployal.


Sendouc:
I don't have a preference :) We do manual deploys anyway so merging often to main is not a problem. Whatever is easiest for you.

PR #1135, opened 2022-11-18
Re-enable bulk edit controls for maplist pools when logged out

remmycat:
I also refactored the `noTitle` prop away by not falling back to a default. Mainly to avoid confusion when e.g. setting `title` and `noTitle` together, and because I think not setting a title is a nicer API for not showing one.


Sendouc (approved):
Nice! And good job getting rid of `noTitle` as well :) That is indeed a lot nicer.

Issue #1141, opened 2022-11-19
Better color for dropdown arrow (light mode)


Sendouc:
![image](https://user-images.githubusercontent.com/38327916/202865838-20b22405-ff61-4fd0-9dc1-af9549c4f1d8.png)

Hard to see it -> easy to not know it's a dropdown. Make it some dark color that looks decent.

PR #1142, opened 2022-11-19
Temporary solution for dropdown arrow in `<select />` elements for Light Mode

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1141

# Description of Changes

Temporary solution for dropdown arrow in `<select />` elements for Light Mode.

I had to use a `.light select {}` CSS selector to do this at the moment, since there are some challenges in passing CSS vars to SVGs if they are used in a `background-image` CSS property. It most likely isn't the best solution, but IMO it's an acceptable one for the time being.

As for the color used, it's just `(0, 0, 0)` with 55% transparency - the exact same analog to the current setting being used for dark mode, `(255, 255, 255)` with 55% transparency

## Screenshot 

image

### DIFF
--- a/app/styles/common.css
+++ b/app/styles/common.css
@@ -269,6 +269,12 @@ select {
   padding-inline: var(--s-3) var(--s-7);
 }
 
+/* Temporary solution for issue: https://github.com/Sendouc/sendou.ink/issues/1141 */
+.light select {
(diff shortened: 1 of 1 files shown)


Sendouc:
I think that's completely fine. Nice job :)

Issue #1146, opened 2022-11-21
Script to generate build analyzer CSV


Sendouc:
Columns different stats
Rows weapons

New script

![image](https://user-images.githubusercontent.com/38327916/203135826-d3db9052-fbec-46aa-aaa4-b0632acc0a26.png)

PR #1148, opened 2022-11-21
Added a script that generates a CSV containing basic attribute information on each Main Weapon

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1146

# Description

Added a script that generates a CSV containing basic attribute information on each Main Weapon


Sendouc:
Yup nice. Then we can extend it if someone else wants something similar :)

Issue #1166, opened 2022-11-25
Feature Request: Weapon icon on build card should link to the builds page that contains all builds corresponding to said weapon


ElementUser:
Feature Request: Weapon icon on build card should link to the builds page that contains all builds corresponding to said weapon

![image](https://user-images.githubusercontent.com/15804376/203894801-ca248b1e-3ad2-408d-8e69-806e2b6b1642.png)

PR #1167, opened 2022-11-25
Feature Implementation: Weapon icon on build card now links to the builds page that contains all builds corresponding to said weapon

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1166

# Description of Changes

Clicking on the weapon icon on a build card now links to the builds page that contains all builds corresponding to said weapon.

Another small change I did was refactor `weaponIdToSlug()`  to a function, `WeaponIdToSlug()`. I then moved it to a more centralized location, since this function is now being used in multiple places.
- I had to capitalize the function name because it uses the `useTranslation` hook, and React hooks require the function name to be capitalized

### DIFF
--- a/app/components/BuildCard.tsx
+++ b/app/components/BuildCard.tsx
@@ -28,7 +28,7 @@ import { EditIcon } from "./icons/Edit";
 import { Image } from "./Image";
 import { Popover } from "./Popover";
 import { InfoIcon } from "./icons/Info";
-import { WeaponIdToSlug } from "~/utils/i18next";
+import { useWeaponIdToSlug } from "~/hooks/useWeaponIdToSlug";
 
 interface BuildProps {
   build: Pick<
@@ -54,6 +54,7 @@ export function BuildCard({ build, owner, canEdit = false }: BuildProps) {
   const { t } = useTranslation(["weapons", "builds", "common"]);
   const { i18n } = useTranslation();
   const isMounted = useIsMounted();
+  const weaponIdSlug = useWeaponIdToSlug;
 
   const {
     id,
@@ -111,7 +112,7 @@ export function BuildCard({ build, owner, canEdit = false }: BuildProps) {
       <div className="build__weapons">
         {weapons.map((weaponSplId) => (
           <div key={weaponSplId} className="build__weapon">
-            <Link to={`${BUILDS_PAGE}/${WeaponIdToSlug(weaponSplId)}`}>
+            <Link to={`${BUILDS_PAGE}/${weaponIdSlug(weaponSplId)}`}>
               <Image
                 path={mainWeaponImageUrl(weaponSplId)}
                 alt={t(`weapons:MAIN_${weaponSplId}` as any)}
--- a/app/components/BuildCard.tsx
+++ b/app/components/BuildCard.tsx
@@ -19,6 +19,7 @@ import {
   modeImageUrl,
   navIconUrl,
   userBuildsPage,
+  WeaponsBuildsPage,
 } from "~/utils/urls";
 import { Ability } from "./Ability";
 import { Button, LinkButton } from "./Button";
@@ -28,7 +29,6 @@ import { EditIcon } from "./icons/Edit";
 import { Image } from "./Image";
 import { Popover } from "./Popover";
 import { InfoIcon } from "./icons/Info";
-import { useWeaponIdToSlug } from "~/hooks/useWeaponIdToSlug";
 
 interface BuildProps {
   build: Pick<
@@ -54,7 +54,6 @@ export function BuildCard({ build, owner, canEdit = false }: BuildProps) {
   const { t } = useTranslation(["weapons", "builds", "common"]);
   const { i18n } = useTranslation();
   const isMounted = useIsMounted();
-  const weaponIdSlug = useWeaponIdToSlug;
 
   const {
     id,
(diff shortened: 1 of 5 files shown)


Sendouc:
This solution seems to be breaking Rules of Hooks ⚠️ https://reactjs.org/docs/hooks-rules.html

Even if it's a bit weird I think you need to refactor it into a custom hook.

ElementUser:
> This solution seems to be breaking Rules of Hooks ⚠️ https://reactjs.org/docs/hooks-rules.html
> 
> Even if it's a bit weird I think you need to refactor it into a custom hook.

Alright, I tried following your recommendation & refactored the code into a custom hook. The feature implementation still works properly - can you verify that the refactor was done properly in my latest commit? Thanks!

Sendouc:
> > This solution seems to be breaking Rules of Hooks ⚠️ https://reactjs.org/docs/hooks-rules.html
> > Even if it's a bit weird I think you need to refactor it into a custom hook.
> 
> Alright, I tried following your recommendation & refactored the code into a custom hook. The feature implementation still works properly - can you verify that the refactor was done properly in my latest commit? Thanks!

This solution still has the same problem I believe.

```ts
const weaponIdSlug = useWeaponIdToSlug;
```

This line is not really doing anything. You are just assigning `useWeaponIdToSlug` to `weaponIdSlug` (not allowed since in react hooks should start with `use`). Then it is having the same problem of calling the hook inside JSX. Try:

```ts
const weaponIdSlug = useWeaponIdToSlug();
```

Then left a nitpick comment to fix but otherwise is looking good and yeah the feature is nice :)

As a side note if you are interesting in React I highly recommend their new beta docs: https://beta.reactjs.org/ It is one of the best docs I have ever read.

app/components/BuildCard.tsx
```diff
@@ -109,13 +112,15 @@ export function BuildCard({ build, owner, canEdit = false }: BuildProps) {
...
         {weapons.map((weaponSplId) => (
           <div key={weaponSplId} className="build__weapon">
-            <Image
-              path={mainWeaponImageUrl(weaponSplId)}
-              alt={t(`weapons:MAIN_${weaponSplId}` as any)}
-              title={t(`weapons:MAIN_${weaponSplId}` as any)}
-              height={36}
-              width={36}
-            />
+            <Link to={`${BUILDS_PAGE}/${weaponIdSlug(weaponSplId)}`}>
```
Sendouc, line 115: Define the value of `to` in urls.ts as a function. I'm trying to keep that file as source of truth for all the links in the project where feasible.

ElementUser:
> > > This solution seems to be breaking Rules of Hooks ⚠️ https://reactjs.org/docs/hooks-rules.html
> > > Even if it's a bit weird I think you need to refactor it into a custom hook.
> > 
> > 
> > Alright, I tried following your recommendation & refactored the code into a custom hook. The feature implementation still works properly - can you verify that the refactor was done properly in my latest commit? Thanks!
> 
> This solution still has the same problem I believe.
> 
> ```ts
> const weaponIdSlug = useWeaponIdToSlug;
> ```
> 
> This line is not really doing anything. You are just assigning `useWeaponIdToSlug` to `weaponIdSlug` (not allowed since in react hooks should start with `use`). Then it is having the same problem of calling the hook inside JSX. Try:
> 
> ```ts
> const weaponIdSlug = useWeaponIdToSlug();
> ```
> 
> Then left a nitpick comment to fix but otherwise is looking good and yeah the feature is nice :)
> 
> As a side note if you are interesting in React I highly recommend their new beta docs: https://beta.reactjs.org/ It is one of the best docs I have ever read.

Yeah I'm not quite as familiar with React Hooks at the moment even after having worked with them once before in a previous PR here (https://github.com/Sendouc/sendou.ink/pull/1070) - before this project, the last time I used React was in 2020, and back then React Hooks were still relatively new (I checked the history & Hooks released in React 16.8 on February 6, 2019). I did not know about the existence of this doc & it explains a lot in its description (that everything in the doc is rewritten with Hooks & not classes, like what I was used to using way back in 2020).

Anyway I'll make the changes, thanks!

app/components/BuildCard.tsx
ElementUser: Okay, I tried doing what you suggested, but by attempting to do so I think I broke the rules of React Hooks again & I'm having a hard time finding a way around this :(

See: https://github.com/Sendouc/sendou.ink/pull/1167/commits/9d5602994f96e74fdbf45193b58f65cac5e7a12c#diff-16131e5525021e86f0194c3c3b3867166462e434cb357fd3198e16d3b964fdd6

Basically, to make the new function in `urls.ts` that gets the weapon build page URL correctly, I need to disguise it as a React component by capitalizing the function name. No matter what, we'll have to end up using `useTranslation()` somewhere to properly render the weapon URL slug, but the problem is that there aren't any real React components in `urls.ts`.

Thing is that I know I shouldn't move this piece of code to something like a `.tsx` file because all we're expecting to get back from this function is text (and not a full-blown React component). The reliance on the `useTranslation()` hook is creating these odd problems & I'm just not seeing a solution that fully satisfies the rules of React hooks at this time, while also satisfying the constraint of centralizing the URL construction in `urls.ts`

ElementUser: Oh, I just thought of something: putting the hook directly into `urls.ts`:

```js
export function useWeaponsBuildsPage(weaponId: MainWeaponId): string {
  const { t } = useTranslation("weapons");

  return mySlugify(t(`MAIN_${weaponId}`, { lng: "en" }));
}
```

Let me know if this is okay to do & I can do another refactor. Thanks again for your feedback!

Sendouc:
I wasn't able to explain it that well so I just went ahead and implemented it so hopefully it's easier to see what I was meaning. Happy to talk about it more if it doesn't really make sense. The way the hooks work sometimes it's just a bit weird and takes time getting used to.

PR #1174, opened 2022-12-01
Object DMG AP UI Relocated and Icons For AP Effects

CardBoardBoss:
The UI for AP has been moved  over the damage receivers that receive its effects


Sendouc:
Great 👍 thanks for contributing.

Issue #1180, opened 2022-12-05
Ability Chunks calculation should ignore Ability Doubler


ElementUser:
I'll fix this when I have the time, just realized that my code doesn't ignore Ability Doubler at the moment for Ability Chunk calculations

PR #1186, opened 2022-12-06
Ability chunk calculation now properly takes Ability Doubler into account

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1180

# Description of Changes

Ability chunk calculations now properly takes Ability Doubler into account
- Ability Doubler ability in of itself does not contribute towards ability chunk count.
- For the Clothing gear, if it has Ability Doubler, then the cost of adding a non-duplicate secondary (sub) ability only costs 3 ability chunks. Reference: https://splatoonwiki.org/wiki/Splatfest_Tee#Splatoon_3_2
- Build Analyzer also no longer shows Ability Doubler as an Ability Chunk entry, nor does it show the Ability Chunks accordion if Ability Doubler is the only selected ability in the build

# Unit Tests

I also added some unit tests to make sure that Ability Chunk calculations are behaving as intended when Ability Doubler is present on the Clothing gear.

Screenshot of passing unit tests for the ability chunks test file:

image

### DIFF
--- a/app/modules/analyzer/abilityChunksCalc.test.ts
+++ b/app/modules/analyzer/abilityChunksCalc.test.ts
@@ -39,6 +39,19 @@ GetAbilityChunksMapAsArray("Empty build results in an empty array", () => {
   assert.equal(abilityChunksArray, [], "Ability chunks array is not empty.");
 });
 
+GetAbilityChunksMapAsArray("Build ignores Ability Doubler", () => {
+  const buildWithOnlyAbilityDoubler = [
+    ["UNKNOWN", "UNKNOWN", "UNKNOWN", "UNKNOWN"],
+    ["AD", "UNKNOWN", "UNKNOWN", "UNKNOWN"],
+    ["UNKNOWN", "UNKNOWN", "UNKNOWN", "UNKNOWN"],
+  ] as unknown as BuildAbilitiesTupleWithUnknown;
+
+  const abilityChunksArray = getAbilityChunksMapAsArray(
+    buildWithOnlyAbilityDoubler
+  );
+  assert.equal(abilityChunksArray, [], "Ability chunks array is not empty.");
+});
+
 GetAbilityChunksMapAsArray(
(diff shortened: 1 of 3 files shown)


Sendouc:
This only came to my mind now, sorry! But actually Ability Doubler behavior is slightly different:

> Adding Abilities to the Splatfest Tee costs less ability chunks as well: adding a different primary ability costs 6 ability chunks, while adding a non-duplicate secondary ability only costs 3.

Source: https://splatoonwiki.org/wiki/Splatfest_Tee#Splatoon_3_2

So I think the analyzer should actually take this in account instead of ignoring it!

ElementUser:
I don't think we should modify the build variable after taking a look, I just have to modify `AbilityChunksRequired()` and `getAbilityChunksMapAsArray()` to accept another parameter that defines whether we're using a Splatfest Tee or not. In the future we can convert this to a type as well & construct the type in the `BuildAnalyzerPage()` component. Anyway.....I might have a look at this another day when I find a chance

Sendouc:
Only Splatfest Tee can have Ability Doubler though so I think it should just take it in account automatically. I'm not a fan of adding more inputs it just makes the UI confusing.

And sure no hurries 👍

ElementUser:
> Only Splatfest Tee can have Ability Doubler though so I think it should just take it in account automatically. I'm not a fan of adding more inputs it just makes the UI confusing.
> 
> And sure no hurries 👍

I can certainly do that & it's definitely a lot easier, but it's moreso if they choose to replace Ability Doubler as the main ability - if they do, everything still costs less on the Splatfest Tee in terms of chunks.

But honestly this is such a small thing that I doubt most players will include Ability Doubler in their build (since it only applies during Splatfests anyway) :D - so yeah I can do what you suggested too whenever I find a chance

ElementUser:
@Sendouc I made a commit that does what you requested =)

Added another unit test to verify the results & edited the original post accordingly to reflect the latest changes

Sendouc:
Cool!

Issue #1193, opened 2022-12-14
Clear weapon inputbox after selecting weapon / before next selection


DoubleCookies:
When choosing a weapon, the input field is not automatically cleared.

DoubleCookies:
Did not find a good solution - the field is not completely cleared

ElementUser:
Honestly I like the way it currently functions =)

DoubleCookies:
Well, _"it just works"_.
I thought that clearing it will be better but it's more difficult than I expected.

ElementUser:
> Well, _"it just works"_. I thought that clearing it will be better but it's more difficult than I expected. Maybe it's not an issue at all?

I don't think it's an issue at all, personally

Sendouc:
If I understand this issue correctly I agree that it's not really a problem. But if we find some cases where it could be awkward UX or cause real problems lets revisit.

PR #1194, opened 2022-12-14
Clear weapon input on focus

DoubleCookies:
#1193 
`clearsInputOnFocus` already works in some other input fields and it's great! I guess how it works is enough for good experience with weapon selector.


Sendouc:
Hmm from my testing it doesn't seem to work. It just clears the underlying value (what weapon is selected) but not the input's value (what user has typed). So maybe the whole prop is a bit bugged but yeah.

DoubleCookies:
I guess it works same way as input in damage calc. I'll carefully check how it works.

Sendouc:
I had so many problems with it before 😵 If it seems hard then dw about it or ask me for help!

DoubleCookies:
I'll open it later if I find good solution

PR #1197, opened 2022-12-23
Add Gear Abilities Article

teddinotteddy:
I'm not exactly sure how adding articles work, but I did try to follow the format that I saw all the other articles were in. The article I added is supposed to help people decide what abilities to use and how to grind for them. I probably made some mistakes, so please let me know if there is anything I should fix or add.


Sendouc:
Did you close it on purpose? :)

Issue #1161, opened 2022-11-23
Dynamic background size


Sendouc:
Currently values are hard coded but it would be nice if it dynamically filled users screen (with min-size set or otherwise it will get really tiny on mobile). That way user wouldn't have to scroll and zoom to make the background fit their screen nicely if they have a larger one.

Feature request:

<img width="621" alt="image" src="https://user-images.githubusercontent.com/38327916/203592366-e74f8cc5-a90b-47f3-8b8c-aed278bd7b42.png">

PR #1200, opened 2022-12-30
Implemented some dynamic background sizing & positioning calculations for the Map Planner

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1161

# Description

Implemented some dynamic background sizing & positioning calculations for the Map Planner.

This won't work well on mobile in portrait mode, since the user won't be able to access most of the controls (honestly, a user should not be attempting to use the feature in portrait mode anyway because it's impractical). It will work fine on landscape without any kind of min-sizing (as mentioned in the original issue).

# Screenshots and Examples

- 24" monitor, full screen (1920 x 1080):

image

- 27" monitor, full screen (2560 x 1440):

image

- Pixel 5 in Landscape Mode (851 x 393):

image

- iPad Mini in Portrait Mode (768 x 1024):

image

### DIFF
--- a/app/hooks/usePlannerBg.ts
+++ b/app/hooks/usePlannerBg.ts
@@ -1,0 +1,34 @@
+import { useState, useEffect } from "react";
+import { useWindowSize } from "react-use";
+
+type PlannerBgParams = {
+  bgWidth: number;
+  bgHeight: number;
+  pointOffsetX: number;
+  pointOffsetY: number;
+};
+
+// Dynamic background size. See this issue for more info: https://github.com/Sendouc/sendou.ink/issues/1161
+const bgSizeFactor = 0.80;
+
+export function usePlannerBg() {
+  const [plannerBgParams, setPlannerBgParams] = useState({});
+
+  // Natively available WindowSize hook: https://usehooks-ts.com/react-hook/use-window-size
+  const windowSize = useWindowSize();
+
+  useEffect(() => {
+    const bgWidth = windowSize.width * bgSizeFactor;
+    const bgHeight = windowSize.height * bgSizeFactor;
+
+    // Point offsets that move the image closer to the center of the window
+    const pointOffsetX = bgWidth * (1 - bgSizeFactor);
+    const pointOffsetY = 0.6 * (bgHeight * (1 - bgSizeFactor)); // Removes some dead space above the image
+
+    setPlannerBgParams({
+      bgWidth, bgHeight, pointOffsetX, pointOffsetY
+    });
+  }, [windowSize.width, windowSize.height]);
+
+  return plannerBgParams as PlannerBgParams;
+}
--- a/app/hooks/usePlannerBg.ts
+++ b/app/hooks/usePlannerBg.ts
@@ -12,7 +12,12 @@ type PlannerBgParams = {
 const bgSizeFactor = 0.8;
 
 export function usePlannerBg() {
-  const [plannerBgParams, setPlannerBgParams] = useState({});
+  const [plannerBgParams, setPlannerBgParams] = useState<PlannerBgParams>({
+    bgWidth: 1200,
+    bgHeight: 800,
+    pointOffsetX: 240,
+    pointOffsetY: 96,
+  });
 
   // Natively available WindowSize hook: https://usehooks-ts.com/react-hook/use-window-size
   const windowSize = useWindowSize();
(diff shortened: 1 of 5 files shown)


ElementUser:
I can probably improve this slightly by moving the background image upwards based on a percentage of whatever percentage of the height the "deadzone" above the image actually takes up. It will just make the point offset calculation slightly more complicated

ElementUser:
Okay, adjusted. Things look good to me now - @Sendouc whenever you get a chance, can you please have a look & see if you can give me any feedback on the changes?

Thanks =)

Sendouc:
Very nice! 💯 🔥  It's so satisfying trying different screen sizes and it just adapts haha.

I made some adjustments so please check. Importing the hook from correct place + a bit more zoomed in version. I think it looks nicer but what do you think?

Sendouc (approved):

ElementUser:
Sure, I'll give it a look when I can!

Also I just noticed, the initial x/y coordinates where it spawns stuff is also hardcoded at like [250, 1000] and [250, 750] in the code - so I might have to spend a bit more time to accommodate that (probably this weekend). It may be better off for me to move some of this stuff to a hook that calls the native Windows hook & also provides some output parameters regarding say, a point of reference for the up-to-date X/Y coordinates of the window (so that I don't have to repeat all the X/Y coordinate transformations from `handleAddBackgroundImage`)

ElementUser:
@Sendouc after looking through the latest changes, I think that a bgSizeFactor of `0.85` is too large - it causes some overlapping UI elements, which don't look too nice - this most notably happens on my full-screen 2560 x 1400 monitor:

image

While it's difficult to fully fix this for most/all resolutions, I do understand that you want the background image to be as large as possible. To compromise, I reduced the bgSizeFactor from `0.85` to `0.80` and it looks much better: 

image

Hopefully you'll accept this change!

ElementUser:
@Sendouc I fixed the image spawn location as well so it is now based on the size & position of the background image (rather than randomizing based on statically defined values).

The "box" for where these images spawn can be approximated by spawning enough images. For example, see the screenshot below:

image

I had a look through the maps & Museum d'Alfonsino has the largest horizontal minimap size (from what I can tell), so my current image spawn box location should be in a good spot.

I do have a question about where to store adjustable constants though - would you like those to be moved somewhere else? I can't think of a good spot for these & I'm not sure if you're okay with leaving the constants within the actual function that spawns an image on the canvas.
-  See this commit for all the details: https://github.com/Sendouc/sendou.ink/pull/1200/commits/9222feeaee255b297552d4ef0dbbf3a46b930306

Also, please let me know if you want me to add/remove comments or elaborate/add clarifications on the math - I tried to not get overly verbose with it, while still trying to say what I'm doing with each of those variables.

Thanks! :)

app/hooks/usePlannerBg.ts
```diff
@@ -0,0 +1,37 @@
...
+  bgHeight: number;
+  pointOffsetX: number;
+  pointOffsetY: number;
+};
+
+// Dynamic background size. See this issue for more info: https://github.com/Sendouc/sendou.ink/issues/1161
+const bgSizeFactor = 0.8;
+
+export function usePlannerBg() {
+  const [plannerBgParams, setPlannerBgParams] = useState({});
```
Sendouc, line 15: I think instead of doing type assertion below we should do:

```suggestion
  const [plannerBgParams, setPlannerBgParams] = useState<PlannerBgParams>({});
```

and give it some reasonable default value. Otherwise the return value of this function is kind of cheating. Or we can ofc:

```suggestion
  const [plannerBgParams, setPlannerBgParams] = useState<PlannerBgParams | null>(null);
```

And let caller handle it.

app/hooks/usePlannerBg.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    setPlannerBgParams({
+      bgWidth,
+      bgHeight,
+      pointOffsetX,
+      pointOffsetY,
+    });
+  }, [windowSize.width, windowSize.height]);
+
+  return plannerBgParams as PlannerBgParams;
```
Sendouc, line 36: This is the type assertion I talked about.

Sendouc:
Yeah that's nice 👍 I left one comment.

app/hooks/usePlannerBg.ts
ElementUser: Thanks Sendou! I knew I was missing something when it came down to the typing for the React Hook.

I went with your first suggestion :) - you can have another look & merge when ready. Glad to see that I did something significant for the map planner!

Sendouc:
Very nice job with this!

PR #1202, opened 2023-01-03
Add an Article: Grinding for Abilities

teddinotteddy:
An article on how to get pure gear. If I made any mistakes please let me know. I tried to follow the same format as all the other articles.


Sendouc:
Hmm I'm not sure how I feel about the gear seed calculator so I don't really want guides detailing it on my site. It's a cool thing and your guide is good so it's not that. Just my personal preference. I'm up to have a guide about grinding gear otherwise.

Issue #1157, opened 2022-11-23
Planner make progress save locally


Sendouc:
It should save locally to IndexexDB. But if you refresh page all progress is lost. Investigate why and fix.

ElementUser:
I checked just now & it's actually not writing to IndexedDB at all (on any web browsers). It always shows up as empty when viewing the Map Planner on the website.

I did check other TLDraw app websites & they write to IndexedDB (I see entries that update whenever I add/remove a shape, but data exists in IndexedDB even without doing anything in particular)

The data for TLDraw on your website must be writing it to somewhere else & we're unable to view where on our web browser - is there a chance that the data could be written on the Remix server side somehow (even if it's temporary)?

Sendouc:
Hmm no I don't think there is any chance of something like that happening. This is something we are doing wrong with TLDraw.

ElementUser:
Just wanted to add some documentation regarding the original issue (copied from Discord): the fix for this primarily involves just adding an `id` field to the `<Tldraw />` component when using it:

![image](https://user-images.githubusercontent.com/15804376/212572998-2558958d-4776-4a0e-8f2f-d203c0934f9d.png)

![image](https://user-images.githubusercontent.com/15804376/212572971-2488a7a8-6065-4c67-96f6-8e67d106e1d4.png)

An example of this can be found from the Tldraw repo: https://github.com/tldraw/tldraw/blob/main/examples/tldraw-example/src/persisted.tsx

ElementUser:
I think this can be closed now 🙂

PR #1207, opened 2023-01-15
Map Planner now persists its state properly by using IndexedDB

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1157

# Description of Changes

Map Planner now persists its state properly by using IndexedDB (via the Tldraw library).

Sendou already submitted the fix after we worked on it via Discord - what I primarily did here was refactor the usage of `<Tldraw | null>` and getting rid of the `if !app return;` statements.

## Other minor changes

Upgraded Tldraw library version from 1.26.4 to 1.27.0.

I tested the functionalities of the current implementation of Tldraw for the Map Planner & everything still works as intended!
- The changelog notes for 1.27.0 can be found here: https://github.com/tldraw/tldraw/commit/55a8831a6b036faae0dfd77d6733a8f585f5ae23#diff-bf0b1a4e9ba4c229ddb424c03b9cb5976744bf1deb0bf2474259f23c67b14654

### DIFF
--- a/app/features/map-planner/components/Planner.tsx
+++ b/app/features/map-planner/components/Planner.tsx
@@ -181,7 +181,11 @@ export default function Planner() {
           {t("common:plans.poweredBy", { name: "tldraw" })}
         </a>
       </div>
-      <Tldraw id="tldraw-persist-sendou" showMultiplayerMenu={false} onMount={handleMount} />
+      <Tldraw
+        id="tldraw-persist-sendou"
+        showMultiplayerMenu={false}
+        onMount={handleMount}
+      />
     </>
   );
 }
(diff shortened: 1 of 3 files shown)


app/features/map-planner/components/Planner.tsx
```diff
@@ -34,7 +29,7 @@ export default function Planner() {
   const { i18n } = useTranslation();
   const plannerBgParams = usePlannerBg();
 
-  const [app, setApp] = React.useState<TldrawApp | null>(null);
+  const [app, setApp] = React.useState<TldrawApp>(new TldrawApp());
```
ElementUser, line 37: I prefer defining an empty default value for our Tldrawapp object here, just so we won't have to deal with `null` related checks if we plan to use the `app` state more often.

Sendouc:
Thanks a lot for helping out with figuring out the root problem it was a good help. I don't really feel comfortable merging this though as I don't quite understand how the change will affect behavior. What I did was copied from some other example and it works so I prefer to leave it as is.

PR #1208, opened 2023-01-19
Updated/added Danish translations

FrederikFraJylland:
Added Danish translations for new lines and made a few adjustment.

For more information look at the comments for the changes of the induvidual json-files.


FrederikFraJylland:
Note.

The "Check translations json" check have failed because the {{count}} variable is missing. in the ""upload.afterExplanation_one": line in the common.json file.

I have not used that variable in the translation of that line, as it is only used, when you need approval of one picture.
In Danish it is more natural to write the smaller numbers as a letter, so I choose to write the number 1 in letter form.

I can change the translation, so it correctly use the variable if it is needed or dessired.

Issue #1265, opened 2023-02-20
Drop Roller AP should be 30 AP per ability, not 10 AP


rmpurp:
On the website Drop Roller applies 10 AP for rsu, ssu and res, when it should be 30 AP (see https://leanny.github.io/splat3/ability.html and click on the drop roller icon or see the video below of a visual comparison of drop roller vs. 10 AP) 

https://user-images.githubusercontent.com/10505811/220000481-1fe1fec4-6d8c-4907-95f3-4fc7fff6be3e.mp4

PR #1266, opened 2023-02-20
Fix Drop Roller AP: 30 AP per ability

rmpurp:
Change DR to give 30 AP per ability: See #1265 

Note that in my testing, DR *does* stack with normal abilities unlike Tacticooler (tested by running >30 AP of ink res alongside DR; as soon as the DR runs out you take ink damage, meaning that it does stack and doesn't simply raise it to 30 if it is less than that)


Sendouc:
Nice catch! Thanks for the contribution 💯

Issue #1276, opened 2023-02-25
Copyright line in footer overlaps the left nav bar (and occupies too much horizontal space in general)


ElementUser:
![image](https://user-images.githubusercontent.com/15804376/221364697-624a8cde-40b7-45a9-a320-cfad95a91968.png)

Easy fix should be to split this into 2 lines

Sendouc:
Nice catch! I have max-width on the patrons list shown in your screenshot but seems like I forgot it from the copyright section.

ElementUser:
Ah, I forgot to mention that for some reason, I use 175% zoom on the site - either way I'll open a PR :) ty Sendou

PR #1277, opened 2023-02-25
Fixed copyright note so that it does not overlap other UI elements horizontally at high zoom levels

ElementUser:
# Linked Issue

https://github.com/Sendouc/sendou.ink/issues/1276

# Description

Fixed copyright note so that it does not overlap other UI elements horizontally at high zoom levels (200% or higher).

Screenshot of the copyright notice after the change (at 200% zoom level):

image


Sendouc:
Nice 👍

Issue #1289, opened 2023-03-02
Side nav icons difficult to discern without labels


justintricate:
![image](https://user-images.githubusercontent.com/47122570/222382290-065b79d0-f8ca-42f4-850d-0b0f206b24f5.png)

It can be difficult at a glance to tell what each icon means at a glance. Could use built in "title" on the <Image> or a more stylish tooltip text. Or neither :P your call, Sendou-- I am submitting a PR with the flyout tooltip text created and styled if you wish!

Sendouc:
Unfortunately had to revert the PR as it was prevent links from being clicked. Saw this issue on badges page but probably there would have been issues elsewhere as well. I think maybe a different approach would be needed.

justintricate:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

MDN talks about using tooltips, which we could implement for sure. It does beg the question-- if the information is important enough for a tooltip, isn't it important enough to always be visible?

Perhaps the icons being re-designed to show their labels is the _actual_ answer, rather than an aesthetic workaround? Worth considering...

Sendouc:
Yeah it's an interesting question! I guess my philosophy was that since they are clearly defined on the front page the side nav is supposed to be little more than "quick access" so lacking some clarity is ok. Although it could be true that when the number of them grow the current solution will soon become obsolete anyway.

Sendouc:
Eventually will be fixed by the whole side nav being gone.

PR #1290, opened 2023-03-02
Sidenav styling

justintricate:
Addresses #1289 

Creates a flyout label for each sidenav icon, using the common:pages.{item.name} so as not to grab "t" or "object-damage-calculator"


justintricate:
it looks like playwright is failing because the tooltip flyout does, in fact, overlap the nav item (they go to the same place)

it doesn't visually overlap, but I'll let you be the call on if you want to use this code or not

image

ElementUser:
Personally I agree on the premise of the issue you're trying to resolve, though the change should be accommodating of as many devices as possible using the principles of responsive design.

On a large screen, showing the text to the right might be okay, but doing so also increases the amount of horizontal screen space each sidenav element is taking up. I personally like the way the icons on the main page are done, with the text being placed **below** each icon:

image

In my opinion, it would be better if we placed some small text below each icon in the sidenav menu.

ElementUser:
Okay just checked on my mobile device & it doesn't even show a side bar, which helps minimize the concern of text being added horizontally. But even on a desktop monitor (24 inches or less anyway), at default zoom levels with the web browser window being maximized, it could still take up too much space horizontally depending on the language used. This could become more of an issue if the user zooms into the page more (for example, I legitimately use a 200% zoom level on my 27 inch monitor).

justintricate:
I think that's a good idea too, either would be good. 😄 I wanted to show this way in case the clean aesthetic of the "icon only" nav was desirable. Putting the nav labels underneath would be easy as well. 

If space is a concern, but the clean look is preferred, could scale the tooltip down some, place it under or over the icon, things like that.

Sendouc:
Woo cool idea I like it! Personally I'm not concerned about lack of space. Playwright I guess doesn't like this:

image

(nav takes full page) but i'm sure there is some solution here. Not sure if it can be real problem or it's just a Playwright problem?

Also for styling I can participate on that. I'll have more time over weekend.

justintricate:
I did introduce "width: 100%;" into the layout_side_nav, let me see if I can fiddle something more appropriate to appease the playwright gods :) that might be what's causing it!

justintricate:
looks good for me now-- Firefox 110.0.1, from the smallest browser size that shows the nav, up to 4k resolution

Sendouc:
https://user-images.githubusercontent.com/38327916/223219706-253e4ed6-45b3-4699-a536-69d8178b2c76.mov

I think should make it so that only clicking the icon or popover (when showing) would cause navigation. Otherwise there is a risk of miss clicks. I can also help looking into this a bit later if needed 👍

justintricate:
let me know if this looks better @Sendouc ! working for me, added some transitions etc. obviously feel free to style how you see fit. 😄

Sendouc:
Very nice!

PR #1318, opened 2023-03-27
Sidenav styling

justintricate:
Fixed the issue that blocked links, instead putting the tooltip text on top of the icon when hovered. I believe this will be a good change for accessibility, as I know I don't remember from the homepage which icon is which personally. Tested at the smallest media query which the nav bar is visible

Animation


Sendouc:
Thanks for taking time to put this together. I took a look but I think I won't be merging this. I think the tooltips shouldn't overlap the icon but I also don't think it maybe makes sense to put more time into this. I'm adding icons faster to the nav than I thought so probably I need to come up with completely new solution soon anyway. Which might mean the nav will get a big redesign at that point.

PR #1329, opened 2023-04-08
Fixes #1311

### LINKED ISSUE #1311: Display "X players participated" when teams are of size 1
image

If all the teams reported are of size 1 then the text should say that `players participated` not that `teams participated`.

zenpk:
This fixes issue #1311 

### How

Using a new JSON field in `calendar.json` called `participatedPlayerCount` to alternatively show "X players participated".

Before rendering `<Results />`, a helper function checks if any team in results contains more than 1 player. If so then render `participatedCount`, else render `participatedPlayerCount`

### Testing

I've tested locally with the dummy data

show players

show players

show teams

show teams

I'm quite new to front-end development. Please let me know if this needs to be improved. Thank you.

### DIFF
--- a/app/routes/calendar/$id/index.tsx
+++ b/app/routes/calendar/$id/index.tsx
@@ -248,6 +248,9 @@ function Results() {
 
   if (!data.results.length) return null;
 
+  // check if all the teams only contain 1 player
+  const isTeam:boolean = ;
+
   return (
     <Section title={t("calendar:results")} className="event__results-section">
       {data.event.participantCount && (
@@ -343,4 +346,9 @@ function Description() {
       </div>
     </Section>
   );
+}
+
+function checkIfIsTeam():boolean{
+
+  return true;
 }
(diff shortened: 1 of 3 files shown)


app/routes/calendar/$id/index.tsx
```diff
@@ -344,3 +350,28 @@ function Description() {
...
+      | string
+      | {
+          id: number;
+          discordName: string;
+          discordDiscriminator: string;
+          discordId: string;
+          discordAvatar: string | null;
+        }
+    )[];
+  }[]
```
Sendouc, line 369: This is also not a big deal but a small TypeScript tip. I believe you can do something like:

```suggestion
  results: {
    players: unknown[];
  }[]
```

It will same you some typing and makes the function more flexible in some use-cases. TypeScript allows you to pass "more" to the function than you define so it helps to be as unspecific as possible when defining function signature.

Sendouc (commented):
Thanks for your contribution! Just left a few TypeScript tips.

app/routes/calendar/$id/index.tsx
zenpk: Thank you so much! I truly appreciate your helpful tips. They are very valuable for a newcomer like me :)

PR #1335, opened 2023-04-11
Fixes #895

### LINKED ISSUE #895: Provide better UX when build cap is hit
Currently add build button just disappers. Should think of something more user friendly. For example add a new text that informs users they have reached the max amount of builds.

zenpk:
Fixes #895 

### How

Added a field in `builds.json` names `reachBuildMaxCount`, which says "You've reached the max amount of builds."

When reaching the limit, the "Add build" button will be disabled and the text will appear. This gives users more information. (though I highly doubt anyone would reach the max count lol)

Initially, I wanted to place the text to the left of the button, but I tried not to add new CSS.

### Test Result

image

image


zenpk:
Closed by accident, sorry.

Sendouc:
Good job! I did one more commit that showcases how you can use the utility CSS classes to achieve what you described.

zenpk:
> Good job! I did one more commit that showcases how you can use the utility CSS classes to achieve what you described.

That is pretty neat. Thanks for teaching me!

PR #1345, opened 2023-04-17
Fixes #839: Allow user to select favorite badge

### LINKED ISSUE #839: Allow user to select favorite badge
image

Allow selecting which badge is big on initial user page load.

I expect this task to go something like this:

- Sort badges on user page by badge id on default.
- Create migration to add "favoriteBadgeId" or something to user table. (FK user id, badge id -> BadgeOwner table)
- Favorite badge can be set / removed from /edit user page.
- If favoriteBadgeId exists show badge with that ID first (preferably done in SQL not TypeScript)
- Good to have: E2E test that tests that favorite badge can be set, has expected result, can be removed and the removal also shows.

## User request

image

zenpk:
Fixes #839 

## Summary

Added a column "favoriteBadgeId" in `User` table, allowing users to select their favorite badge from the edit page.

### Backend

1. Database change is as described above \(from the migration file\). I didn't add FKs because `BadgeOwner` is now a view rather than a table. If a user's "favoriteBadgeId"=0 \(default\), then it means they haven't specified their favorite badge
2. Added a seed function, to randomly choose a favorite badge for user Sendou
3. Added "favoriteBadgeId"  to `updateProfile.sql`
4. When loading the user's data, sort the badge by id, and put the favorite badge at the top, all done in SQL \(`countsByUserId.sql`\)

### Frontend

1. Added a select box on the edit page, which looks similar to the country select box. The initial value is the user's current favorite badge (if doesn't have one, then is "-")

Initially, I used a ComboBox component, which brings the benefit of displaying the badge image. But how it works is super unintuitive so I replaced it with the select box. Users can choose the "-" option to remove their favorite badge selection.

2. Added favoriteBadgeId validation to the zod schema. **Here comes a problem**: 

For other form data such as country or weapon, they all have a global variable that stores the dataset. But for badge data, the badges owned by the current user are stored in `parentRouteData.badges`, which have to be accessed from the `useMatch` hook or passed by other functions. Neither way is feasible so I just check if the selected badge's id >= 0, which is not very ideal.

### E2E Test Result

1. Select the favorite badge, it will be displayed first, other badges are sorted by id

image

image

2. Unselect the favorite badge, all badges are sorted by id

image

image


Sendouc:
Good work with this 🚀 LGTM!

PR #1353, opened 2023-04-21
Round sub velocity to thousandths place, fix bug in round function (Fixes #1348)

### LINKED ISSUE #1348: Velocity: multiply by ten or show 3 decimals
image

pm-wobbuffet:
Adjusts stats.ts and creates a new case for ability type, "SUB_VELOCITY", which would allow for separate processing in the future and keeps me from accidentally changing other values.

Fixes a bug in the roundToNDecimalPlaces routine in number.ts, where the previous implementation had a hard coded 2 decimal cutoff from the use of 100 instead of 10**n.

First time doing a pull request or using this javascript framework, so feel free to offer guidance or feedback.


Sendouc:
Awesome! Nice job catching the bug with that function as well 👏

Issue #1343, opened 2023-04-16
`meta` (Remix V2 refactoring)


Sendouc:
⚠️ REMIX FUTURE CHANGE: The route `meta` export signature is changing in v2You can prepare for this change at your convenience with the `v2_meta` future flag. For instructions on making this change see https://remix.run/docs/en/v1.15.0/pages/v2#meta

PR #1359, opened 2023-04-26
Refactor: Remix V2 `meta`

zenpk:
Refactors #1343 

### Changes

1. Opt-in to the Remix V2 meta by adding

```javascript
  future: {
    v2_meta: true,
  },
```

to `remix.config.js`

2. Move some `meta` fields into HTML `<meta>` tags, as suggested in https://remix.run/docs/en/1.15.0/route/meta-v2#global-meta

3. Manually search for `MetaFunction` and change them to `V2_MetaFunction` (I actually have no idea if there's a smarter way to do this...)

### Breaking change

According to the official doc, in V1, parent routes' meta data will be merged into the children's (if conflicted then keep the children's). But in V2, Remix will only keep the deepest meta data. So as for now, if `root.tsx` contains some meta data that children's meta data doesn't include, these will be lost (actually, there's only one, which is the "description"). The change is shown below.

#### V1

image

#### V2 (missing "description")

image

It can be solved by adding `{ matches }` to the `V2_MetaFunction`, like this (code from the official doc):

```javascript
export const meta: V2_MetaFunction = ({ matches }) => {
  let parentMeta = matches
    .flatMap((match) => match.meta ?? [])
    .filter((meta) => !("title" in meta));
  return [...parentMeta, { title: "Projects" }];
};
```

In this way, we must explicitly filter every conflict field, which is quite difficult to maintain. And we have to add this to every V2_MetaFunction, which seems a bit messy. So I didn't do it.

### DIFF
--- a/app/features/build-analyzer/routes/analyzer.tsx
+++ b/app/features/build-analyzer/routes/analyzer.tsx
@@ -1,4 +1,4 @@
-import { type LinksFunction, type MetaFunction } from "@remix-run/node";
+import { type LinksFunction, type V2_MetaFunction } from "@remix-run/node";
 import type { ShouldRevalidateFunction } from "@remix-run/react";
 import { Link } from "@remix-run/react";
 import * as React from "react";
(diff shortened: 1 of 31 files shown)


Sendouc:
Good job! And yeah about the breaking change you described probably the right way to deal it is for me to write some descriptions for each page. Better for SEO anyway than using default one everywhere.

PR #1377, opened 2023-05-24
added hebrew translations pt.1

shachar700:
Added translations in Hebrew. Would need to verify that all sentences appear correctly and are not split to parts due to the valued parameters.


Sendouc:
I added some things and now if you run it locally, should be able to select it.

image

Does that look correct?

shachar700:
This looks correct. Most of the translations are fine. It looks great. I think it's ready for public. I will fix the translations that are a bit off/mistyped.

The name for the language should be without diacritical signs (the dots and lines that appear below the letters  = עברית)
A suggestion for the list of languages is to have their names in English next to them so that the alphabet order would make sense for the viewers.
for example "Hebrew - עברית"

shachar700:
Adjusted some translations

PR #1386, opened 2023-06-06
Fix the Sidenav hidden issue

### LINKED ISSUE #1305: Sidenav hidden behind header when screen is short
image

You can't scroll up to see the full nav. Probably some padding is missing?

zenpk:
_**Updated solution in the comment below**_

---

Fixes #1305 

## The problems

There are two problems causing this issue, one is that the header's height needs to be calculated and depends on the `var(--item-size)`, and the other is the flex is justified center so the nav links will overflow the top of the flex box.

To solve them, for the first problem, I set a top property for the sidenav which matches the height of the header. For the second problem, it can be solved by changing to `justify-content: flex-start`.

## Side effects

In order to access the `var(--item-size)` in the `.layout_side-nav` class, we can put the variable in the `:root`. However, to keep it scoped, I created another class called `.layout__item_size` and applied it to both the header and the sidenav, which is a little ugly but works.


zenpk:
I think I missed the point of `justify-content: center` since my laptop has a very small screen. I'll reopen this if I could come up with a better solution.

zenpk:
## The problem with `justify-content: center`

Using `justify-content: center` can cause an unscrollable overflow. In the horizontal flex-direction, we can use `margin: auto` to solve this, but in the vertical direction, it is quite tricky. A newer syntax is `justify-content: safe center` but it is currently only supported by Firefox.

## Solution

I added some ugly client-side JavaScript to detect the window `resize` event, on every resize, check if the first link's `offsetTop` is smaller than the header's `height`, if true then change to `justify-content: flex-start`.

In order to prevent the infinite loop (because after we change to flex-start, it won't overflow anymore so the function will run again to change it back), I added another `useEffect` to double-check the overflow.

## Result

It will cause some bouncing as the overflow state is not changed back and forth in two different `useEffect`s, but considering resizing a window is not a frequent action, I think it is bearable.

Sendouc:
Interesting! I had no idea that was the cause. Thanks for investigating this issue, people frequently comments on it.

So I guess we have two options right:

1) Align the menu from the top with a CSS only solution
2) Keep the current styling but additional JavaScript is required.

I'm kind of thinking that maybe even changing the styling is not the worst (so we could go with 1) ) but what is your personal opinion, which looks better?

zenpk:
@Sendouc As a small-screen user, I didn't notice that it is vertically centered even after visiting the site hundreds of times lol. And I think with more and more icons, the difference will be more subtle, so I'm with option 1 as well. Also, the required JS code is actually pretty ugly in my opinion lol.

zenpk:
@Sendouc No problem! It is now ready to merge.

Sendouc:
Ye very nice!

PR #1413, opened 2023-07-09
Add Chinese Translation

gellneko:
- Added (hopefully) all the missing Chinese translations
- Revised a few translations for consistency


Sendouc:
🙌 '

PR #1416, opened 2023-07-11
Add Chinese Translation

gellneko:
- Added all the missing translations (for real)
- Revised some translations for better accuracy


Sendouc:
🙌

PR #1417, opened 2023-07-11
Add Korean translation for analyzer and game_misc

niLPotential:
I couldn't figure out what turret was in analyzer.json so I just skipped it for now.
Otherwise all lables are checked and translated.


Sendouc:
Hey good job 👍 

Turred is when you dodge roll with Glooga Dualies and the shots concentrate. So that mode before you move again deals increased damage.

PR #1422, opened 2023-07-16
Fix: page scrolls to top when opening dialog

### LINKED ISSUE #1411: Opening Dialog scrolls page to top
For example:

http://localhost:5800/u/nzap/art

To reproduce

1) scroll down a bit
2) click art to preview (open a dialog)
3) close dialog
4) Notice that you are back at the top of the page instead of your previous position

zenpk:
Fixes #1411

After some searching, I was very lucky to find the easiest solution, which is described in this comment:

> PS Omitting the height: 100% effectively locks the scrolling at the current position without the jumping - on latest Chrome, at least. – [zerm](https://stackoverflow.com/users/359335/zerm)
 
https://stackoverflow.com/questions/3656592/how-to-programmatically-disable-page-scrolling-with-jquery

(I actually have no idea why it behaves like that...)

P.S. There's also a JavaScript solution but is not even close as neat: https://stackoverflow.com/questions/4770025/how-to-disable-scrolling-temporarily

### DIFF
--- a/app/styles/common.css
+++ b/app/styles/common.css
@@ -449,6 +449,7 @@ dialog::backdrop {
   border: none;
   margin: auto 0;
 }
+
 /* stylelint-enable no-descending-specificity */
 
 .input-container > input:focus-within {
@@ -772,6 +773,7 @@ dialog::backdrop {
 
 .lock-scroll {
   overflow: hidden;
+  height: unset;
 }
 
 /* https://stackoverflow.com/questions/50917016/make-a-hidden-field-required/50917245#comment117565184_50917245 */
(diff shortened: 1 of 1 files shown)


Sendouc:
Hey! Do you get this kind of behavior on your machine as well? I see the background jumping a bit horizontally on dialog open. It's a small thing but I wonder what is causing it 🤔 

https://github.com/Sendouc/sendou.ink/assets/38327916/fb73cb8f-2218-4b03-814a-ddfd43243f83

zenpk:
@Sendouc I just tested and yes, I got the same behavior. It is caused by the absence of the scrollbar. So the "cause chain" is: `overflow: hidden` -> scrollbar hidden -> display moved horizontally

PR #1425, opened 2023-07-18
Add ko translation for builds, common and others

niLPotential:
Also adjusted some prior translations based on feedback
I omitted parts which do not directly impact the usability of ko users
I don't think I will be translating calender, teams and such for the same reason
I'll come around again when new features are added
Keep up the great work!


Sendouc:
Cool thanks!

PR #1428, opened 2023-07-24
Fixed #1407

### LINKED ISSUE #1407: Not possible to scroll results highlight table on mobile all the way
Located here http://localhost:5800/u/sendou/results

https://github.com/Sendouc/sendou.ink/assets/38327916/fb4f3c01-4166-42c0-bc98-b884ae563042

Notice how the table doesn't scroll all the way to left.

hfcRed:
First time creating a pull request, hope this is alright.

Single line fix, the style "items-center" was applied to the whole table, causing all the content to align to center, which caused overflow if it didnt fit the screen.

The style already gets applied to the individual list items, so removing it from the table caused no further issues.

### DIFF
--- a/app/routes/u.$identifier/results/index.tsx
+++ b/app/routes/u.$identifier/results/index.tsx
@@ -34,7 +34,7 @@ export default function UserResultsPage() {
       {showHighlightsSection && (
         <Section
           title={t("results.highlights")}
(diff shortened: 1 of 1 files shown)


Sendouc:
Thanks for the fix!

PR #1469, opened 2023-08-29
Translation 20230808

aisuzuki:
rough translation for all missing items
(auto translation for json items that copilot does is pure "wow" )


Sendouc:
Thanks a lot again :)

PR #1523, opened 2023-10-15
Damage Calculator Grid Rework

hfcRed:
I have some proposed changes for the damage calculator grid, particularly to make it more mobile friendly.

Currently you have to scroll up/down and left/right a lot, quickly loosing track of the relevant damage type and damage receiver.

opera_IQvssTQKqS

On Desktop it can also sometimes not be obvious that there is more content to the right until you scroll all the way down.

opera_wUpmjlGeG8

By sacrificing a tiny amount of space, its possible to fit all row content on Desktop.
The first row is made sticky, so the relevant information never leaves the screen.

opera_5LUR2QR5wu

On mobile the first column is also made sticky.
This way you can always see which data corresponds to which damage type and damage receiver.

opera_yUjzpJb2sK

I tried to keep all the styling as close to the original as possible, but had to cut down on a few paddings.

Let me know what you think!

### DIFF
--- a/app/features/object-damage-calculator/calculator.css
+++ b/app/features/object-damage-calculator/calculator.css
@@ -25,12 +25,40 @@
 }
 
 .object-damage__grid {
+  column-gap: var(--s-2);
   display: grid;
+  max-height: 60vh;
+  overflow: auto;
+  row-gap: var(--s-2);
   width: 100%;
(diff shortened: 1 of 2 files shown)


Sendouc:
Nice contribution! You are right this works out much better :)

PR #1524, opened 2023-10-16
Page layout change proposal

hfcRed:
I noticed on the VODs page that the content clips into the sidebar before it reaches the breakpoint.

opera_ntHC5It9FD

I was going to make a quick fix by adding some padding, but noticed the reason for this is that both navbar and sidebar do not take any actual space on the page due to having a fixed position.

By moving around some elements and changing them to be sticky its possible to have the same behavior while having the elements actually take up space.

This makes layouting easier and more consistent and avoids clipping. Both navbar and sidebar would also be more future proof to any changes since they can no longer mess with any page content.

It also has the added benefit of making the interaction with the footer a lot more natural!

opera_55FEJnINem

If you look at the main content div you can see that it now respects the sidebar and navbar

Untitled

I made the pull request a draft because in order to avoid clipping in the current layout, all pages have top paddings applied to them, which means I would have to go through every page and adjust the paddings to get the original look back, so I would like to get your thoughts on this first!

### DIFF
--- a/app/components/layout/SideNav.tsx
+++ b/app/components/layout/SideNav.tsx
@@ -12,7 +12,7 @@ export function SideNav() {
       {navItems.map((item) => {
         return (
           <Link
-            to={item.url}
+            to={`/${item.url}`}
             key={item.name}
             prefetch={item.prefetch ? "render" : undefined}
           >
--- a/app/components/Main.tsx
+++ b/app/components/Main.tsx
@@ -1,9 +1,10 @@
-import { useCatch } from "@remix-run/react";
+import { useCatch, useLocation } from "@remix-run/react";
 import clsx from "clsx";
 import type * as React from "react";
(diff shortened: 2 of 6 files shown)


hfcRed:
Awesome, thank you! I will adjust the paddings and then open the PR

app/components/layout/SideNav.tsx
```diff
@@ -12,7 +12,7 @@ export function SideNav() {
       {navItems.map((item) => {
         return (
           <Link
-            to={item.url}
+            to={`/${item.url}`}
```
Sendouc, line 15: Yeah I noticed these relative URL's have bene causing issues... Wonder what's the root cause, maybe Remix bug?

Sendouc:
Very nice job! And good attention to detail with the scroll bar in this and the object damage calc pr :)

PR #1528, opened 2023-10-21
Fix VoDs page weapon display problem

### LINKED ISSUE #1298: VoDs page only showing weapon filtered for and not all in the vod
image

So it should show all weapons in the VoD, not just the one we have selected in the filter.

zenpk:
Fix #1298 

Hi Sendou, it's been a while since the last PR :)

## Fix

The problem lies in the SQL `and vm."stageId" = coalesce(@stageId, vm."stageId")`, which filters out all the weapons that don't match with the query. I moved the logic to JavaScript since I couldn't find a better SQL to achieve the safe effect as `includes()`.

## Results

image

image


Sendouc:
Hello there! :)

yes this seems like an appropriate fix here. Nice one 👌

Issue #1513, opened 2023-09-25
UI Issues - Arrows on the wrong side when selecting Hebrew


shachar700:
In art at the bottom:
![Screenshot_88](https://github.com/Sendouc/sendou.ink/assets/48739719/61bed184-7162-40a5-8c9d-f6e153ca877c)

In menus such as map lists, or in max/min menu of DMG calc:
![Screenshot_89](https://github.com/Sendouc/sendou.ink/assets/48739719/f3ed9875-6315-4f70-85b0-b85b71415108)

as well as in map planner, map lists. 

The toggle in art for open submissions is also a bit off in its location.

duplicatedframe:
I believe this is solved simply by drawing the left arrow on the right button and vice-versa if the user language's direction is `rtl`, as we'd still expect the behavior of the buttons and pagination dots to reverse alongside the language.

Working on implementing and will create a PR shortly.

PR #1531, opened 2023-10-29
improve rtl handling in pagination and front

duplicatedframe:
This closes issue #1513
It sets `icon` to the left arrow on the right button and vice-versa if the user language's direction is `rtl`, as we'd still expect the behavior of the buttons and pagination dots to reverse alongside the language.

Additionally, it styles the class `.front__drawings` to display in Left to Right no matter the direction of the user's language.

### DIFF
--- a/app/components/Pagination.tsx
+++ b/app/components/Pagination.tsx
@@ -3,6 +3,7 @@ import { Button } from "~/components/Button";
 import { ArrowLeftIcon } from "~/components/icons/ArrowLeft";
 import { ArrowRightIcon } from "~/components/icons/ArrowRight";
 import { nullFilledArray } from "~/utils/arrays";
+import { useTranslation } from "~/hooks/useTranslation";
 
 export function Pagination({
   currentPage,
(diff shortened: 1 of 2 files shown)


Sendouc:
Hey thanks for the PR :)

I'm wondering if this is something we should handle in the Javascript and not CSS? Would something like https://developer.mozilla.org/en-US/docs/Web/CSS/:dir work and then just additional styles? e.g. with arrows just rotate them 180deg.

My logic here is that if possible styling related concerns should stay in .css files. What do you think?

hfcRed:
``dir();`` unfortunately has super bad coverage at only 16% of users. All the popular browsers do not support it, or need the user to specifically enable it either through settings or extensions.

opera_TDe0rqfmkn

You could try doing something like this

Add the ``dir`` attribute to the HTML

```html
<div dir="rtl">
</div>
```

Then use an attribute selector in CSS

```css
[dir="rtl"] {

}
```
This would make it more flexible to other situations as well, but now youre back in TSX land.
It does not seem like you can escape doing this in code if you want full browser coverage.

Sendouc:
Hmm ok yeah oops missed that. That said I think that solution might be preferable because then you have the "code solution" in one place and are able to use stylesheets for the rest.

I suspect this might not be the last place where you need to tweak the styles to fit RTL and I'd prefer not to spread that concern around the Javascript more than necessary if possible 🤔

hfcRed:
Found a super simple zero code solution :)
Do you want me to make a new PR?
firefox_CuiKuJLQfT

duplicatedframe:
While I agree with all the above, I'd argue the solution in this gif isn't behaving as expected. In a RTL languge, the pagination should too be displayed RTL.

Thanks for all the support on this!

hfcRed:
That is true! I realized that after making the comment as well and have adjusted the CSS to keep the RTL pagination

duplicatedframe:
Though, crossing that solution with the previous suggestion of rotating the arrows could likely work.

The div containing the arrows and dots could be rotated 180 deg.

hfcRed:
Yep, made a pull request already!
https://github.com/Sendouc/sendou.ink/pull/1535

PR #1535, opened 2023-11-09
Added rtl utility class

hfcRed:
This fixes issue https://github.com/Sendouc/sendou.ink/issues/1513

Since you already applied the ``dir="rtl"`` attribute to the HTML element, we can use a CSS attribute selector to apply a style to a class only if that attribute is present.

This makes for a really simple and clean implementation. If any future RTL issues come up, you can just make your own utility class using the ``html[dir="rtl"]`` attribute selector.


Sendouc:
💯

PR #1536, opened 2023-11-10
default english option now listed as American English (en-US); British English (en-GB) added as an option

SporeCloud:
see title for what this does. the large number of commits is me forgetting how to github.
there is probably a more efficient way to implement this translation. english names for weapons, gear etc. are the same between between the NTSC and PAL versions of Splatoon 2 and 3 as far as I know, so the only real difference between the en-GB and en-US translations really comes from the spelling of certain words, of which there aren't too many.
there are differences between the NTSC and PAL versions of Splatoon 1 though, so if support is added for those (I'm not sure why or when it would be, aside from the scene being kept alive via systems like Pretendo) then this translation will stand out a bit more.


Sendouc:
Hey thank you for the pull request.

I thought about it but I don't think currently there is enough benefit when weighted against the added effort for me when developing the site. Honestly I don't want to have to think about keeping this translation up to date since i'm definitely not an expert about the differences between en-GB and en-US.

PR #1622, opened 2024-01-14
Volume slider

hfcRed:
Just a simple volume slider for SQ sounds

https://github.com/Sendouc/sendou.ink/assets/101019309/d64ff875-fe02-42b3-84f4-7ad8d4c7011e

Should work on all major browsers, cant test Safari unfortunately. Firefox will fire the sound twice when testing for mobile in a desktop browser, but will work correctly when using on an actual mobile device (its always Firefox).

### DIFF
--- a/app/features/sendouq-settings/routes/q.settings.tsx
+++ b/app/features/sendouq-settings/routes/q.settings.tsx
@@ -665,12 +665,11 @@ function SoundCheckboxes() {
 
 function SoundSlider() {
   const [volume, setVolume] = useState(() => {
-    if (isBrowser()) return soundVolume();
-    return 1;
+    return soundVolume();
   });
 
   useEffect(() => {
-    if (isBrowser()) setVolume(soundVolume());
+    setVolume(soundVolume());
   }, []);
 
   const saveVolume = (event: React.MouseEvent<HTMLInputElement> | React.TouchEvent<HTMLInputElement>) => {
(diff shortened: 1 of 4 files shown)


app/features/sendouq-settings/routes/q.settings.tsx
```diff
@@ -659,3 +662,46 @@ function SoundCheckboxes() {
...
 }
+
+function SoundSlider() {
+  const [volume, setVolume] = useState(() => {
+    return soundVolume();
+  });
+
+  useEffect(() => {
+    setVolume(soundVolume());
+  }, []);
```
Sendouc, line 673: This combined with `defaultValue` does not look quite correct 🤔 I tried removing useEffect and I see no change in functionality. But then again updating state is pointless if you are using defaultValue afaik https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable

I would suggest:

- use null (or 100 doesn't matter) as default value in useState
- `defaultValue` -> `value`

Sendouc:
Neat little feature! Check the comment. Also check the CI it's failing (you can also try `npm run cf:noe2e`) locally

app/features/sendouq-settings/routes/q.settings.tsx
hfcRed: I implemented useEffect because of the isBrowser check initially, but youre right that it is no longer needed!
Will refactor to use value instead

Sendouc:
Nice work!

PR #1718, opened 2024-04-16
Add recycled brella damage

AKonicki26:


AKonicki26:
Adds damage calculations for the recycled Brella to the object damage table. The name of the brella as well as the damage calculations are taken from @lobotomysyndrome on discord as of 03/01/2024.

PR #1719, opened 2024-04-16
Add recycled brella damage

AKonicki26:
Adds damage calculations for the recycled Brella to the object damage table. The name of the brella as well as the damage calculations are taken from @lobotomysyndrome on discord as of 03/01/2024.

(Closing the first PR and creating this one to squash commits)

### DIFF
--- a/scripts/create-object-dmg-json.ts
+++ b/scripts/create-object-dmg-json.ts
@@ -45,6 +45,8 @@ for (const cell of Object.values(params.CellList)) {
   if (!DAMAGE_RECEIVERS.includes(cell.ColumnKey)) continue;
   if (!cell.DamageRate) continue;
 
+  //console.log(cell);
+
   if (!result[cell.RowKey]) {
     result[cell.RowKey] = {
       mainWeaponIds: weaponParamsToWeaponIds(weapons, cell.RowKey).filter(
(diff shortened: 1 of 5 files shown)


Sendouc:
Hey there. This object-dmg.json should be generated by the `create-object-dmg-json.ts` script and not be modified by hand. So from a quick look I think what is missing is that either:

1) Adjust the script generating the json so that it generates it in correct shape without manual adjustments
2) Change the object dmg calc code so that it reads splat brella multipliers for recycled.

Probably 1) is better

AKonicki26:
Done! Wasn't aware that those builders existed. Thank you much for the feedback! I'm not sure if it's possible to be doing the checks without the hardcoded values in there, but based on the way that they're already used in a few places, I don't believe its a bad idea here

scripts/create-object-dmg-json.ts
```diff
@@ -45,6 +45,8 @@ for (const cell of Object.values(params.CellList)) {
   if (!DAMAGE_RECEIVERS.includes(cell.ColumnKey)) continue;
   if (!cell.DamageRate) continue;
 
+  //console.log(cell);
+
```
Sendouc, line 49: Should be removed

Sendouc:
Hmm I don't still get the same json generated by the command. Run the command (`npx tsx scripts/create-object-dmg-json.ts`) to generate the json then compare the one in the scripts/output folder to the one you have commited here. They seem different 🤔

Sendouc:
Ok nm my bad checked something wrong. Yep this looks good! Thanks for the contribution :)

PR #1725, opened 2024-05-07
Adds missing badge info + some rewording

Ant-Spl:
Info summary for visitors:
- Added missing info about comissioning borzoic for a badge with a rough estimate of the cost in BRL;
- Rolled back the translation "conjunto" back to "build" in the analyzer as no one here actually says that;
- Did you know that "build" is gendered as a female word in Brazilian Portuguese? Now the gendered terms should be correct in the "builds" tab;
- Changed "localidade" to "região" in tournament restrictions to make it make sense;
- We all know events can happen in Salmon Run and in Tableturf Battles but it should be clear in the wording that the event itself is Salmon Run or Tableturf based;
- "Top Search" now formely translated as "Pesquisar Tops" has been reworded to "Listas dos Top 500" to make it clear that the page itself is a list of top 500 players sorted by division and season and not a Google Search style page. Now some users can stop looking for a search bar in the page;
- Added gendered option indicator to captain in the teams tab. My apologies to all Brazilian she/her captains!;
- "forms.showDiscordUniqueName.info" asked a question to the user but, the page layout doesn't really indicate the need of a question. I swapped the question with a description.
- A few grammar fixes, especially prepositions and hyphens.

Thank you very much!


Sendouc:
💯

PR #1737, opened 2024-05-22
Fix broken link to Locales folder

ximk:
Noticed the link was broken while looking at the translations section of the README, so here is the fix for that.

**Before:**
image

**After:**
image


Sendouc:
Thanks!

Issue #1733, opened 2024-05-19
Persist LFG filters state to search params


Sendouc:
https://github.com/Sendouc/sendou.ink/blob/rewrite/app/features/lfg/routes/lfg.tsx#L53 in essence make this backed by search params. This makes it so that filter state can be shared by sharing the URL and refreshing the page doesn't lose this state.

References:

- `useSearchParamState.ts`
- `builds.$slug.ts` (a bit different since filtering is done on server-side but here should be done on client side)

What needs to be solved:

- Filter state is potentially huge, is it ok just to throw it as "JSON.stringify" to the URL?
- Should we validate filter state when it's coming out of URL? Probably not too important.

PR #1739, opened 2024-05-24
Added support for sharing a filter set using the URL

KaiserOfNone:
This completes https://github.com/Sendouc/sendou.ink/issues/1733
Also added support for custom encoder for `useSearchParamState`, this should enable us to avoid having a big json blob on the url bar :p

Demonstration of the code working: https://testing.avillagra.net/lfg?q=l.en-w.0%2C1-mn.1

### DIFF
--- a/app/features/lfg/lfg-types.ts
+++ b/app/features/lfg/lfg-types.ts
@@ -121,7 +121,7 @@ export function smallStrToFilter(s: string): LFGFilter | null {
       if (!languagesUnified.some((lang) => lang.code === val)) return null;
       return {
         _tag: "Language",
-        language: val, // Kinda trusting the language I get is good, bad idea.
+        language: val,
       };
     }
     case "pt": {
--- a/app/features/lfg/lfg-types.ts
+++ b/app/features/lfg/lfg-types.ts
@@ -1,6 +1,8 @@
-import type { LFGType } from "~/db/tables";
+import { LFG_TYPES, type LFGType } from "~/db/tables";
 import type { MainWeaponId } from "~/modules/in-game-lists";
 import { TIERS, type TierName } from "../mmr/mmr-constants";
+import { assertUnreachable } from "~/utils/types";
+import { languagesUnified } from "~/modules/i18n/config";
 
 export type LFGFilter =
   | WeaponFilter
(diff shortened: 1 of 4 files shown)


app/features/lfg/lfg-types.ts
```diff
@@ -45,3 +45,109 @@ type MinTierFilter = {
...
+      return `tz.${filter.maxHourDifference}`;
+    case "Language":
+      return `l.${filter.language}`;
+    case "PlusTier":
+      return `pt.${filter.tier}`;
+    case "MaxTier":
+      return `mx.${tierToNum.get(filter.tier)}`;
+    case "MinTier":
+      return `mn.${tierToNum.get(filter.tier)}`;
+  }
```
Sendouc, line 88: We can use `assertUnreachable` in the default case here to ensure at a build time all the cases have been handled

app/features/lfg/lfg-types.ts
```diff
@@ -45,3 +45,109 @@ type MinTierFilter = {
   _tag: "MinTier";
   tier: TierName;
 };
+
+const typeToNum: Map<LFGType, number> = new Map([
+  ["PLAYER_FOR_TEAM", 0],
+  ["TEAM_FOR_PLAYER", 1],
+  ["TEAM_FOR_COACH", 2],
+  ["COACH_FOR_TEAM", 3],
+]);
```
Sendouc, line 54: What happens if a new `LFGType` is added? IIRC If we use `Record<LFGType, number>` here instead of a map it will check for us a build time we are handling all the cases.

app/features/lfg/lfg-types.ts
```diff
@@ -45,3 +45,109 @@ type MinTierFilter = {
...
+      if (Number.isNaN(n)) return null;
+      return {
+        _tag: "Timezone",
+        maxHourDifference: n,
+      };
+    }
+    case "l": {
+      return {
+        _tag: "Language",
+        language: val, // Kinda trusting the language I get is good, bad idea.
```
Sendouc, line 124: I guess we can do `languagesUnified.some(language => ...` above and then it's for sure it's part of the language selection 🤔

Sendouc (commented):
Very clean solution! It will be definitely nice to have this feature.

I left some comments for consideration :)

app/features/lfg/lfg-types.ts
KaiserOfNone: Yeah I guess adding an array called `LFGTYPES` like for `TIERS` would work. Both cases are similar, for the tiers I'm hoping whenever a new one is added the array is updated. Didn't find a way to iterate over all cases of a type sadly, that'd be a nice ts feature :(

app/features/lfg/lfg-types.ts
KaiserOfNone: I found not having a return outside the switch statement does the same thing implicitly, but I like the explicitness of your option.

Sendouc:
🚀

Issue #1741, opened 2024-05-25
Tournament embed logo preview to work on Twitter


Sendouc:
<img width="594" alt="image" src="https://github.com/Sendouc/sendou.ink/assets/38327916/c5ba48b2-92c1-444b-84a7-3458a19aa678">

What would it take to get the logo to show up here as well? Discord embed works great

KaiserOfNone:
I think we need to add some twitter specific tags to tell it how to render it maybe [1], I'll experiment with it later

[1]: https://developer.x.com/en/docs/twitter-for-websites/cards/guides/getting-started

PR #1744, opened 2024-05-26
Added twitter's special tags

KaiserOfNone:
Added Twitter's special tags. They were required for rendering because they are special.

Their docs don't mention anything about renaming `name="twitter:card"` into `name="x:tag"`, let's hope Elon doesn't notice.

Fixes https://github.com/Sendouc/sendou.ink/issues/1741

Screenshot 2024-05-26 at 9 03 53 AM


Sendouc:
Quite interesting they use their own custom tags but good to know :)

Issue #1755, opened 2024-06-03
Well Stringer & Decavitator object damage


hfcRed:
![image](https://github.com/Sendouc/sendou.ink/assets/101019309/f2895787-ecd0-474b-9ed3-da5d037675cc)

Damage multipliers for both Well Stringer and Decavitator do not seem to exist, calculator falls back to defaults. It doesnt seem like the generated damage JSON generated the parameters for the weapons.

PR #1756, opened 2024-06-04
Fix Wellstring and Decav damage numbers

ashbinary:
Object damage was redone incorrectly, missing Wellstring and Decav and showing incorrect numbers. Addresses #1755


Sendouc:
Thanks for this!

PR #1759, opened 2024-06-07
Add new weapon alt names

ximk:
I could try going further and adding in more specific abbreviations (ex. vunder and sunder), but I feel like that would be overkill. This commit just adds/"fixes" the ones that stand out.


Sendouc:
Very nice!

> can we makle order splatana replica show up when you type stamper btw

From Discord ☝️

ximk:
Can do!

ximk:
Fixed (alongside the other missing Side Order kits) in https://github.com/Sendouc/sendou.ink/pull/1759/commits/8403ce808c3d2b24b628468c6c9ec0ce7dc7731c.

Issue #1159, opened 2022-11-23
Undoing map bg image set should display previous bg image if any


Sendouc:
Now it just goes blank:

https://user-images.githubusercontent.com/38327916/203585461-af81f98c-29c9-4c89-8b78-464d8e2e7864.mov

ElementUser:
This is caused by `app.resetDocument();` in `app\features\map-planner\components\Planner.tsx`.

Further investigating would be required to see if this function call is needed & re-evaluate from there. I have a few thoughts though (I haven't read the docs yet):

1. If the function call is needed, we would have to manually save a snapshot of the canvas history before calling resetDocument()
2. If the function call is **not** needed, then we should be able to simply remove it - but it's important to know exactly what this function does & why it was here in the first place

ElementUser:
At this time, a good way to do this may be to remove the reset document call on the "Set background" button dg implement a separate reset button instead. But the screen is already full of buttons, so UI-wise I don't know if that is desirable.

I do like the "Set background" button's current behaviour of actually clearing out the other stuff on the canvas though - so do we really still consider this a valid bug when it currently flows nicely UX-wise (IMO)?

Sendouc:
Yeah I think it's still an issue because when you think what you expect "undo" to do then resetting to previous state is that. + you might accidentally blow away some important progress that currently can't be restored.

Issue #1158, opened 2022-11-23
Clicking on background image should make weapon image lose focus


Sendouc:
Currently double click is needed. It works correctly before you add the background image.

https://user-images.githubusercontent.com/38327916/203584990-e7374707-0014-4402-a14a-e1b4673f8d96.mov

ElementUser:
I notice that hovering the cursor over the background image still marks it as "selected" (and thus, it still grabs the focus of the cursor), despite it being locked. I'll look through the TLDraw documentation to see if there's a way to make it not able to grab the focus of the cursor.

ElementUser:
So I spent some time & I don't think they have an API method to make the background image ignore your cursor, so to speak.

I noticed that you can still right-click the background image to unlock it if you really wanted to.

Sendouc:
Yeah it's not a problem if they unlock it with a right click. The problem is just the UX of when you want to lose focus. Currently it doesn't really work how you'd expect. Meaning if in any drawing program you click the background you expect to lose focus of the current thing.

PR #1762, opened 2024-06-16
Update TLdraw to 2.2.0

### LINKED ISSUE #1745: Update tldraw to latest
@tldraw/tldraw (powers map planner) latest version right now is 2.1.4

Afaik there are some API changes so might be good to work on this separately from usual dependency upgrading thing.

kytoaa:
Updated TLdraw to 2.2.0 (latest version) in the map planner, had to do some editing to the html and css as there were a few bugs with TLdraw stuff behind the weapon and map selectors.

Fixes #1745 

Screenshot 2024-06-16 221551


kytoaa:
Sorry didn't notice it wasn't passing some checks, can't fix right now but will do as soon as I can

kytoaa:
Alright fixed the issue, I forgot to run prettier, all checks passing!

Sendouc:
Looking good! Believe this also Closes #1158

Only thing I noticed is that it seems that behavior changed when you replace one background image with another. Previously it cleared the last background image but in your branch it stays behind. It's not a huge deal but what kind of work do you think that'd be to fix?

kytoaa:
Fixed! Also decided to fix issue #1159 while I was at it, only needed 1 extra line!

Sendouc:
Awesome! Great work

Issue #1764, opened 2024-06-17
Add French (NoA localization) as a language


ximk:
![image](https://github.com/Sendouc/sendou.ink/assets/104683822/851974d5-1b3d-40a7-8cbb-19c2e24c5f93)

Example (with the Zebrafish Hi-Tops):
![image](https://github.com/Sendouc/sendou.ink/assets/104683822/ea400121-cb3c-48fc-b25d-9934c702efd9)

The plan here should probably be to copy the French translations already present on the site and add it as a new language, and then edit the files `game-misc.json`, `gear.json`, and `weapons.json` to reflect the NoA localizations.

I've translated `game-misc.json` already, I'll just need the auto-generated files for the NoA localization. I'll submit what I have in a PR shortly.

Sendouc:
Let me know if you can't figure out how the auto-generation works. Tbh I kinda forgot myself but I can find out

ximk:
Got it to work! :)

Later, I'll PR a change to the README with a more detailed instructions for the monthly update.

PR #1766, opened 2024-06-18
Add French (NoA) Localization

ximk:
The new folder `fr-CA` copies everything from the current French translations and only changes localization differences.


ximk:
Addresses #1764

Sendouc:
Cool! Still need to add them to `languages` and `languagesUnified`in modules/i18n/config.ts

Not 100% sure what the display name should be but maybe you have an idea.

ximk:
These are what I settled on, as it emphasizes that it's just localization differences. We could do Français (Americas) and Français (Europe), but I think this is fine.

French (NoA)
image

French (NoE)
image

PR #1767, opened 2024-06-18
Added missing keys for JA translation

yutarour:
I have added all of the keys that were mentioned in https://github.com/Sendouc/sendou.ink/issues/1104#issue-1437510715 for Japanese.


Sendouc:
Thank you! I'm sure this will of use to many people.

PR #1771, opened 2024-06-23
Add 1v1, 2v2, and 3v3 Tags

ximk:
Straightforward, just adds tags for 1v1, 2v2, and 3v3 tournaments so players can find them easier (goes along with https://github.com/Sendouc/sendou.ink/commit/c5a051a1e163ce0e942352c550bbd8a94472731a).

Would it be possible to automatically set these if the tournament is not a 4v4 tournament?


Sendouc:
I think it's a good idea and should be automatic however it might be a good idea to refactor the "automatic tags" logic a bit not to make it so brittle to do this sort of thing.

Sendouc:
Idk why I thought this was an issue lol but yeah adding them just as manual ones is fine for now as well.

Issue #1769, opened 2024-06-23
Allow looking for individual coaching


Sendouc:
New LFG post type

![image](https://github.com/Sendouc/sendou.ink/assets/38327916/09c2833f-b16e-4656-b664-c49ac687d76c)
https://discord.com/channels/299182152161951744/1251247392603246814/1251247392603246814

ximk:
Would just using the current "looking for team" template be enough? I can PR this if so:
![image](https://github.com/Sendouc/sendou.ink/assets/104683822/7008e44c-cd61-4dba-bbc4-02114acee1f8)
![image](https://github.com/Sendouc/sendou.ink/assets/104683822/06507806-850a-4489-98ee-eaee23d26990)

Sendouc:
Yes I think that works

PR #1775, opened 2024-06-24
Add "looking for individual coaching" and "looking for scrim partners" options

ximk:
Very simple change, just adds a new LFG type.

Addresses #1769.

### DIFF
--- a/app/db/tables.ts
+++ b/app/db/tables.ts
@@ -271,6 +271,7 @@ export type LFGType =
 	| "PLAYER_FOR_COACH"
 	| "TEAM_FOR_PLAYER"
 	| "TEAM_FOR_COACH"
+	| "TEAM_FOR_SCRIM"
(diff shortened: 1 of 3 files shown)


ximk:
@Sendouc Can you look into why this typecheck is failing? I've been searching for possible reasons for a bit and can't come up with a fix.

Sendouc:
You need to modify the `LFGType` as defined in tables.ts :)

I would also get rid of this redundant constant on the file you modified. It should use the defined right under `LFGType` instead (`LFG_TYPES`)

ximk:
This new error should be easier to fix, however I am lost again. 😅

ximk:
Typecheck now passes thanks to @cysabi. :)

ximk:
I'm going to add a "Looking for scrim partner" option as well, as I think that would be useful.

PR #1789, opened 2024-06-29
Add "equals" option to build filter

ximk:
As requested in the Discord:
image


Sendouc:
Hmm honestly not convinced it's worth it to add an extra element there. To me it's quite niche use-case and you can already do it by selecting both as the person requesting pointed out.

PR #1814, opened 2024-07-29
Add an option to sort by public or private builds on a user's own builds page

### LINKED ISSUE #1783: "Private first" & "Private last" build sorting options
image

rikipls:
# Fixes #1783 

Added Public Build and Private Build as an option to sort builds by (of course, the effect of doing this is only visible to a user's own build page since private builds are inherently invisible to other users)

Unit tests added:
- [app/features/user-page/core/build-sorting.server.test.ts](https://github.com/Sendouc/sendou.ink/compare/rewrite...rikipls:sendou.ink:rewrite?expand=1#diff-64808727f9f8430367083f17813a32e5c9deab5593eda78c6c83fa394e2f5cec)
  - Tests the two sorting options both individually and together (using only one will do the job, but order will matter when using both)


Sendouc:
Nice one :)

PR #1826, opened 2024-08-10
Add Triple Splashdown fist as receiver to object dmg calc

### LINKED ISSUE #1656: Add triple splashdown hand damage receiver data
See also https://github.com/Sendouc/sendou.ink/issues/1396

ximk:
Closes #1656

### DIFF
--- a/app/features/object-damage-calculator/core/object-dmg.json
+++ b/app/features/object-damage-calculator/core/object-dmg.json
@@ -1,4521 +1,4919 @@
 {
-	"Blaster_BlasterLightShort": {
-		"mainWeaponIds": [230, 231],
-		"subWeaponIds": [],
-		"specialWeaponIds": [],
-		"rates": [
-			{
-				"target": "BulletUmbrellaCanopyCompact",
-				"rate": 1.6
-			},
-			{
-				"target": "BulletUmbrellaCanopyNormal",
-				"rate": 2.2
-			},
-			{
-				"target": "BulletShelterCanopyFocus",
-				"rate": 2.2
-			},
-			{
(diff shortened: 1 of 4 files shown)


ximk:
Converting this PR into a draft as I've now learned TSD is affected by OS, and in the meantime I am going to ask dataminers for those multipliers.

ximk:
Damage multiplier for OS has been added, should be good to go now.

app/features/object-damage-calculator/core/object-dmg.json
Sendouc, line 1: I should document this better but this file should not be updated manually but instead generated by the `/create-object-dmg-json.ts`.

So we need to adjust the file above to generate the kind of `object-dmg.json` we want.

ximk:
Working on formatting issues now, object-dmg.json is now generating correctly. Only thing now is that the formatting is off for some reason.

Sendouc:
Oh yeah you can just run `biome:fix` script in the package.json to fix that.

So it's like content needs to come from the script but formatting comes from Biome.. :D yeah could be better

ximk:
Fixed everything, should be good to go now!

image

Sendouc:
Sorry took a long time to take a look but just one final question: where does the 100hp come from?

Otherwise looks good 💯

ximk:
> Sorry took a long time to take a look but just one final question: where does the 100hp come from?

https://github.com/Leanny/splat3/blob/2e6337a95e146a89515ca380a0c850e786236c2a/data/parameter/810/weapon/WeaponSpPogo.game__GameParameterTable.json#L104

Sendouc:
Ok cool! Thanks, good addition.

PR #1835, opened 2024-08-13
Remove outdated script example

### LINKED ISSUE #1834: Update README to provide current method of updating FC via a script
The script `update-fc.ts` no longer exists, so the command [here](https://github.com/Sendouc/sendou.ink?tab=readme-ov-file#update-friend-code) needs to be updated.

ximk:
Closes #1834

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,12 +165,6 @@ Some common files:
 
 ## Commands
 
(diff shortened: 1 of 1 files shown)


Sendouc:
Ty!

PR #1841, opened 2024-08-21
Updated Danish Translations

FrederikFraJylland:
The first pull request I made from this failed on the checks due to me using double spaces instead of tabs.
I tried to change it, but I still got failueres on the check.

So I closed the original pull request and made it again, where all the files I have changes starts every line with a tab instead of double space. Hopefully this will mean the JSON-files are now formatted correctly


Sendouc:
Thanks!

Issue #1806, opened 2024-07-19
Show currently selected value(s) on the ability curve chart


Sendouc:
![image](https://github.com/user-attachments/assets/8081b0aa-ce58-4825-93c1-8db085dab791)

One or both (in case of comparison)

When hovering should work as now, before hovering just show the current values.

Adding the lines as suggested in the image above can be a nice bonus.

PR #1848, opened 2024-08-24
Added cursor line rendering to the Build Analyzer chart and fixed .env.example

ElementUser:
Issue Link: https://github.com/Sendouc/sendou.ink/issues/1806

# Description of Changes

Added cursor line rendering to the Build Analyzer chart when a data point is selected (default value for this option is `false` when not specified)
- Also fixed an "Invalid URL" error when following the README instructions & running the application locally (by adding the `BASE_URL` field to `.env.example`)

==

This is how the Build Analyzer chart looks like after the change when hovering over the chart:

image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,6 +25,8 @@ TWITCH_CLIENT_SECRET=
 SKALOP_SYSTEM_MESSAGE_URL=http://localhost:5900/system
 SKALOP_TOKEN=secret
 
+BASE_URL=https://example.com
+
 VITE_SITE_DOMAIN=http://localhost:5173
 VITE_SKALOP_WS_URL=ws://localhost:5900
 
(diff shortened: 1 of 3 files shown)


ElementUser:
@Sendouc unfortunately, I don't know how to fix the original issue that the reporter reported in the linked issue (https://github.com/Sendouc/sendou.ink/issues/1806), but I thought I would add the cursor lines since that was one of their requested "nice-to-have" features

Sendouc:
Hmm I'm not convinced this is an improvement unfortunately. The issue you linked is specifically about plotting the current value. I disabled those lines on hover on the purpose as I find it's redundant with the tooltip.

About BASE_URL that's a valid point if following README instructions does not result in a working env! I will make another issue to explain a bit better what is wrong there (this value should be removed completely)

Sendouc:
https://github.com/Sendouc/sendou.ink/issues/1850 here

Issue #1774, opened 2024-06-24
Enable Biome rule "noSvgWithoutTitle"


Sendouc:
1) Enable noSvgWithoutTitle by removing the line in biome.json
2) Fix the problems

ElementUser:
This can be closed now =)

PR #1849, opened 2024-08-25
Enable Biome rule "noSvgWithoutTitle"

ElementUser:
Issue Link: https://github.com/Sendouc/sendou.ink/issues/1774

# Description of Changes

Enabled `noSvgWithoutTitle` by removing the line in `biome.json`
- Added the `<title />` tags in all the icon files
- Note that there is no localization provided - when I tried, `biome.json` still threw errors, unfortunately. Not to mention that there are a ton of localizations to go through (I went through 1/3 of it before I realized the linter still complained about the lack of `<title />`  even though I put them into conditional render tags


Sendouc:
Yep it's fine to skip the i18n here imo. Thanks!

Issue #1851, opened 2024-08-25
Upgrade @headlessui/react to latest


Sendouc:
Some API changes.

Should do manual validation on the UI elements that it still works the same after upgrade (as well as fixing the type errors etc.)

ElementUser:
Left a comment in my MR, but copy-pasting it here just in case:

> So I took a look at the @headlessui/react Repo and it is a confirmed issue with the version of the library itself. Someone else reported a bug that is very similar to this (very likely the same issue): [tailwindlabs/headlessui#3396](https://github.com/tailwindlabs/headlessui/issues/3396)
> 
> So it's best to not proceed with the upgrade until they fix it in a later version. However, we can keep an eye on this linked Github issue at least =)

PR #1857, opened 2024-08-25
Upgraded @headlessui/react to latest version (2.1.3)

ElementUser:
Issue Link: https://github.com/Sendouc/sendou.ink/issues/1851

# Description of Changes

Upgraded @headlessui/react to latest version (2.1.3)
- Refactored many components of the form `Component.Subcomponent` (which are marked as deprecated by VSCode) to the separate/more direct imports (e.g. `Combobox.Input` -> `ComboboxInput`)

Tested the Combobox search feature in Build Analyzer and User Search & general navigation around the website, and I did not catch any functionality that broke.

===

**Note:** will require a rebuild & reupload of the bun lock file after merging other Pull Requests that contain dependency changes.


Sendouc:
https://github.com/user-attachments/assets/c972508e-4a99-44d9-874c-a01bd28dac96

Hmm otherwise I think looks good but this layout shift is a bit unfortunate.

ElementUser:
Very subtle change that you caught, thanks for double-checking Sendou!

So I took a look at the @headlessui/react Repo and it is a confirmed issue with the version of the library itself. Someone else reported a bug that is very similar to this (very likely the same issue): https://github.com/tailwindlabs/headlessui/issues/3396

So it's best to not proceed with the upgrade until they fix it in a later version. However, we can keep an eye on this linked Github issue at least =)

PR #1859, opened 2024-08-25
Updated README.md for instructions to run locally

ElementUser:
Updated the "Running locally" section of README.md with the correct bin usage (bun)
- Added installation step for the tsx bin & updated the installation doc URL (previous URL does not work anymore)


Sendouc:
Thanks, the culprit was my fast copy paste.

```
Prerequisites:
- Install latest version of [tsx](https://tsx.is/getting-started)
```

We can drop this as `npx` will install it automatically if you don't have it.

ElementUser:
> Thanks, the culprit was my fast copy paste.
> 
> ```
> Prerequisites:
> - Install latest version of [tsx](https://tsx.is/getting-started)
> ```
> 
> We can drop this as `npx` will install it automatically if you don't have it.

Done =)

package.json
```diff
@@ -105,8 +105,11 @@
 	},
 	"trustedDependencies": [
 		"@biomejs/biome",
+		"@swc/core",
```
ElementUser, line 108: I'm not sure where these come from, but they aren't in the main branch - I'll work on removing them

ElementUser: Looks proper now =)

Issue #1853, opened 2024-08-25
Upgrade @tldraw/tldraw to latest


Sendouc:
I get this error when trying to set background image after upgrade:

![image](https://github.com/user-attachments/assets/3744aaf8-b52c-469b-aa9f-06c7ac223107)

Are we doing something wrong?

ElementUser:
@Sendouc I found the issue =)

I was reading through the tldraw 2.4.0 release notes (since it was the most recent major release since 2.3.x) & going through each point one-by-one and seeing if your application has anything from there: https://github.com/tldraw/tldraw/releases/tag/v2.4.0

Turns out that it does!

> The fileSize property of TLImageAsset and TLVideoAsset is now optional. 

I removed that field from the appropriate line & tested it locally, the issue is fixed. I will submit a Pull Request very soon! =)

PR #1860, opened 2024-08-25
Upgraded @tldraw to latest version (2.4.5)

ElementUser:
Issue Link: https://github.com/Sendouc/sendou.ink/issues/1853

# Description of Changes

Upgraded @tldraw to latest version (2.4.5)
- One of the fixes in the 2.4.0 major release (https://github.com/tldraw/tldraw/releases/tag/v2.4.0) made the `fileSize` field optional in `TLImageAsset`
- By extension, it also means the value of that field has to be non-zero. I simply removed the field & it fixed the bug encountered in the issue link

===

**Note:** will require a rebuild & reupload of the bun lock file after merging other Pull Requests that contain dependency changes.


Sendouc:
Good find! thank you

Issue #1852, opened 2024-08-25
Upgrade remix-auth-oauth2 to latest


Sendouc:
Some API changes

Note that for this you need to set up the auth for your local environment.

PR #1861, opened 2024-08-25
Upgraded remix-auth-oauth2 to latest version (2.3.0)

ElementUser:
Issue Link: https://github.com/Sendouc/sendou.ink/issues/1852

# Description of Changes

Upgraded remix-auth-oauth2 to latest version (2.3.0)
- Refactored OAuth2Strategy constructor fields to the ones mentioned in the official documentation: https://github.com/sergiodxa/remix-auth-oauth2?tab=readme-ov-file#directly

===

Note: will require a rebuild & reupload of the bun lock file after merging other Pull Requests that contain dependency changes.


Sendouc:
Hmm seems like something else might be needed

image

ElementUser:
> Hmm seems like something else might be needed
> 
> image

Thanks for testing @Sendouc!

I did some issue lookups in the `remix-auth-oauth2` repo & found this: https://github.com/sergiodxa/remix-auth-oauth2/issues/94

Since 2.1.0, the default authentication strategy was set to `request_body`. I took a look at the Typing for the `OAuth2StrategyOptions` options based on the linked issue above, and found this in the index.ts file:

```ts
/**
     * The method to use to authenticate with the Identity Provider, this can be
     * either `http_basic_auth` or `request_body`.
     * @default "request_body"
     */
    authenticateWith?: "http_basic_auth" | "request_body";
```

So `request_body` is the default auth method selected - does Discord/your app use `http_basic_auth`? If so, we can try adding in `authenticateWith: "http_basic_auth"` and see if that fixes it. Otherwise, I don't know what else to do at the moment unless I can get the auth flow set up locally

PR #1864, opened 2024-08-29
Added more danish translations

FrederikFraJylland:
Added more danish translations for the tournament.json file and other files


Sendouc:
Thank you!

PR #1874, opened 2024-09-04
Translation Fixes (Spanish Primarily)

Mario64iscool2:
Added all missing Latino Spanish keys. Didn't do much with Spain Spanish because I'm not nearly as confident with it.
Enforced the site name when written in text as 'sendou.ink' unless at the beginning of a sentence.


Mario64iscool2:
The linter is now happy, so this should be ready to push to prod whenever.

Sendouc:
Cool thanks!

PR #1879, opened 2024-09-09
Update the A41 logo to its latest version

ximk:


Sendouc:
Ah thanks but these icons are legacy and won't be updated.

PR #1895, opened 2024-09-29
Add badge for multiple tournaments

ximk:
PRing with permission from owners (see individual commits).


Sendouc:
Seems like public/static-assets/badges/cookie.png is animated when it should be static

ximk:
Fixed in latest commit.

PR #1899, opened 2024-10-01
OCE Open badge by SSKarot (Oceanic)

SSKarot:
hi sendou, hope im doing this right cuz this is literally my first time using git hub 😭🙏

if "OCE Open" isn't allowed as a name, could "Oceanic Open" be allowed? im just hosting tournaments for oceanic regions only tis all


Sendouc:
Seems like this PR does not contain the actual badge files.

SSKarot:
thats really strange, i also just realised i messed up on something with the avif file so id have to send another pull request to make sure its all correct. ill also try to fix the badge files not showing up itself, thank you and i apologise.

SSKarot:
the badges should be added now, if they still aren't there then i honestly have zero clue unfortunately sorry.
image

PR #1929, opened 2024-10-13
Add Rocket Science Badge

RileyStonesifer:
I committed the file add and tournament description separately, but otherwise followed the guide in the badge channel as close as possible.


Sendouc:
Hello avif should be static (not animated)

PR #1933, opened 2024-10-14
Add Heaven 2 Ocean Qualifier badge

ricebowls3:


ricebowls3:
Added the Seabattle badge as well

PR #1934, opened 2024-10-16
Add Tristrike Tuesdays: All Modes (Silver Bracket) badge

marcushatesshot:
not sure if it's supposed to have all my previous commits on it too but i couldnt figure out how to get rid of them soo


Sendouc:
Sorry this is too different from the existing badges so it won't be added.

PR #1936, opened 2024-10-18
Craft Company Badges

mCodestatic:
Re-synced with main branch, fixed AVIFs to be static, & rewrote homemade.ts change to match


Sendouc:
The rotation speed seems too low

mCodestatic:
I didn't change it at all from default, I'm not sure how to on the web viewer. I'll edit it though

PR #1937, opened 2024-10-18
Add Minus Badges

jgiefer:


Sendouc:
The rotation speed seems too low

jgiefer:
Increased the rotation speed :D

Sendouc:
Looks good now but the .avif file needs to be static. mb didn't see in the last go through

jgiefer:
avif files are static now

PR #1949, opened 2024-10-29
Add Squeaky Clean Scrap and Chi-Shoals badges

ricebowls3:


ricebowls3:
Added Shark Tank: Deep Waters badge

PR #1957, opened 2024-11-08
Add Get Coupled badge

wooperissy:


wooperissy:
This was with permission of the badge artist, attempted to add in the last PR but it was too slow. this is the edited version

PR #1974, opened 2024-11-29
Added German Translations for tournament.json

jgiefer:
Accidentally named the commit "made minus badge gifs larger", gifs are unchanged though. Only translation is updated


Sendouc:
thank you!

PR #1977, opened 2024-12-03
Add SpaceDog Showdown and SpaceDog Showdown (Jolly Edition) badges

ricebowls3:


ricebowls3:
Added Badge Palooza badge

PR #1978, opened 2024-12-04
Added ja translations for missing keys.

yutarour:
It's me again from #1767. 
Came around and saw some missing keys in JA (#1104). Went ahead and translated them. Please let me know if there are any issues.


Sendouc:
thank you a lot!

PR #1986, opened 2024-12-15
Added "Fifty Caliber League" First, Second, and Third place badges

CryoniKi:


Sendouc:
I feel like the rotation speed is a bit too low on these when you compare to e.g. https://github.com/Sendouc/sendou.ink/pull/1984/files

PR #1991, opened 2024-12-20
Add Golden Brush

DreamyVeemo:
lmk if theirs any promblems!


Sendouc:
The .avif file should be static

PR #1996, opened 2024-12-24
Splashdown Saturday badge :)

Terrrett:
This is requesting to create the badge for Splashdown Saturday by the organization Ammo Knights. :)


Sendouc:
- .avif should be static 
- rotation speed is too slow (it should take about 3 sec to go the full round)

PR #1997, opened 2024-12-25
plz add saladbowl badge

ItIsSkitty:
someone else might request to have access to this badge but that'll happen at a later point. thank you very much!


Sendouc:
the avif should be static

PR #2011, opened 2024-12-31
HatsuneMiku and Pink NES badge

DreamyVeemo:
commesioned for new ink lan


Sendouc:
The rotation speed for the NES should be higher (about 3 sec for full round). Otherwise looks good.

PR #2012, opened 2025-01-02
add Plus 12 Badge

Kliffbar:
creating a badge for plus 12 tournaments


Sendouc:
The files don't seem correct (end in ".lnk")

Kliffbar:
I fixed the file extension please take another look at it when you can. 
Thanks.

Sendouc:
Some problems:

- too slow (should be around 3 sec for full round)
- static preview should be one frame of the animation
- should be vertically & horizontally centered
- avif is not the same as png
- avif has the picocad logo

PR #2016, opened 2025-01-04
Reef-Lux Achievement and Meet Your Match badges (hopefully truly fixed)

Terrrett:


Sendouc:
There are some problems. At least:

- the rotation should be counterclockwise
- they should be centered vertically
- static image should be a freeze frame from the gif (now it's "zoomed out")

PR #2017, opened 2025-01-04
Add Mesozoic Mayhem Badges (5)

Akainacephalus:
Add the following badges for Mesozoic Mayhem:
Mesozoic Mayhem: Triceratrials
Mesozoic Mayhem: Gold Triceratrials
Mesozoic Mayhem: Special Spinosaurus
Mesozoic Mayhem: Gold Special Spinosaurus
Mesozoic Mayhem: Dreadnoughtus Undefeated

### DIFF
--- a/app/components/elements/DatePicker.tsx
+++ b/app/components/elements/DatePicker.tsx
@@ -1,0 +1,82 @@
+import clsx from "clsx";
+import {
+	Button,
+	Calendar,
+	CalendarCell,
+	CalendarGrid,
+	DateInput,
+	type DatePickerProps,
+	DateSegment,
+	type DateValue,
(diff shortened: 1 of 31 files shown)


Sendouc:
A few things:

- also update homemade.ts
- static image should be from same angle (probably from the side)
- they might also be off-center? at least vertically

Akainacephalus:
i think i fixed the issues but im not sure

Sendouc:
I think for consistency it should still be adjusted. Here is what I see at least:

- "Tricera(gold)" should be more zoomed in. If it's that small it won't look good on the site
- "Deadnoughtus" has too much space below. ("Spino(gold)" too)

PR #2018, opened 2025-01-05
Add Catgirl Challenge badge

wooperissy:
adding on behalf of marcus and Kattail


Sendouc:
The rotation speed is too low. Should be about 3 sec for the full round.

wooperissy:
Thanks, should be fixed now

PR #2020, opened 2025-01-06
Added Low Tier Experiments Badge

Enderchase64:
I missed a comma in the first pull request for this.


Sendouc:
Hmm it has some consistency problems compared to existing badges. Mainly you can see through it (the different blocks aren't fully connected) and also the rotation seems a bit off? Not sure why maybe just because it's not upright.

PR #2034, opened 2025-01-15
adding literallyyyy the best badge

SSKarot:
Screenshot 2024-12-21 193924


Sendouc:
Rotation should be counterclockwise

PR #2039, opened 2025-01-18
StarZ Impact Badge

Ilumixxx:


Sendouc:
- avif should be static
- png is weirdly pixelated
- not centered
- gif looks weird, probably not taken how instructed..?
- too slow rotation

PR #2053, opened 2025-01-23
Soul League badges

YuriHanaa:
Hi, these are badges for Soul League hosted by Splatos!


Sendouc:
Hey looks good otherwise but the .avif file should be static not animated

PR #2057, opened 2025-01-26
Add netherportal badge

Taysonbucket:


Taysonbucket:
please add I fixed it and madure there are files

Sendouc:
- avif should be static
- it rotates too slow
- some graphical glitching in the gif
- too zoomed in

PR #2060, opened 2025-01-26
Soul League Badges fixed please add

YuriHanaa:
I fixed the avif file to be static


Sendouc:
- the revocup one should be a bit more zoomed in. Otherwise it won't look good on the site
- the rotation should be counterclockwise

PR #2073, opened 2025-02-02
Added 4000 XP Badge

Spritzu:


Sendouc:
These are added a bit differently. I will check what we will do with these then maybe put it up myself. But thank you in any case I like the design :)

PR #2080, opened 2025-02-06
Add Barnacle Bash badge

Yumpy42:


Sendouc:
I don't think the text banner should disappear in the gif. Possibly should also be zoomed in just a bit more?

Sendouc:
Sorry but it still doesn't look right. The text should not teleport like that.

PR #2081, opened 2025-02-06
Adds "A Splatoon Brazil Tournament" badge

Ant-Spl:
This badge will be given to winners of tournaments hosted by the Splatoon Brazil staff.

We hold tournaments only 2 to 3 times a year and they all follow this "national" theme.

We chose a Brazilian flag as our model, as it highlights our presence in Splatoon. One side has Inkling eyes and the other Octoling eyes.

Thanks in advance.


Sendouc:
- avif should be static
- full rotation should take around 3 sec
- should be centered (both vertical and horizontal)
- should be more zoomed in

Ant-Spl:
@Sendouc I did the changes. I hope it's okay now.

Issue #1544, opened 2023-12-02
Tri Stringer damage vs. Booyah wrong


Sendouc:
![image](https://github.com/Sendouc/sendou.ink/assets/38327916/8f27884f-cdba-47dd-87e0-3b34fc0b4e97)

https://github.com/Sendouc/sendou.ink/assets/38327916/f1e83665-ee38-446a-a36e-f3c0ccb7318e

Might need some special case?

ipaulhamus:
Hello, I am working on a solution to this issue. I was looking through the "help wanted" tag and noticed that the Tri-Stringer damage on Booyah is still 2 HTD on the site.

PR #2090, opened 2025-02-10
Stringer Max Booyah Damage Fix

ipaulhamus:
Proposed solution to Issue #1544 

Added a special case to the "otherDamage()" function that returns 0 if the weapon is Tri-Stringer and the special is Booyah. Let me know if there are any changes I should make to the implementation, thanks!

To test, run the development build, navigate to the Damage Calculator, and select Tri-Stringer/Inkline Tri-Stringer. The intended result is 136.5 DMG and 4 HTD in the "maximum" damage type, with the others being unaffected.


Sendouc:
Looks good! Thanks

PR #2103, opened 2025-02-20
Add Coffee Cup and Tayson Gauntlet Badges

Taysonbucket:


Sendouc:
This zones concept badge already exists https://sendou.ink/badges/2

Taysonbucket:
> This zones concept badge already exists https://sendou.ink/badges/2

If I removed the zones badges would you be able to still add the Coffee Cup Badge?

PR #2104, opened 2025-02-22
Add Birthday Edition Badge

ItIsSkitty:
my birthday is on march 1st, if this could be done before then thay would be greatly appreciated!


Sendouc:
Where are the gifs from? I think the fps is higher than picocad web viewer or something

ItIsSkitty:
The gifs are from web viewer, no change to the fps

Sendouc:
Maybe my eye was just in higher fps earlier

PR #2113, opened 2025-02-24
Add New Game + Badge

CarlosRev124:


Sendouc:
- res is too high compared to existing badges
- rotation too slow
- should be more zoomed in

CarlosRev124:
updated files sent

PR #2116, opened 2025-02-27
Badge for Bingo Tournament

AstroDDev:


Sendouc:
- not sure what this is made with but they look too high res compared to existing badges
- not centered (vertically)

AstroDDev:
The badges guide recommends 512x512 so that's the resolution I rendered it out in on the picoCAD viewer. I'll make sure to rerender the badge and center it

PR #2118, opened 2025-02-28
update birthday edition badge

ItIsSkitty:
making it not look so high quality, I'd do the same for the ferris wheel badge but it would look really bad if I did


Sendouc:
Don't change the fileName it should remain the same

ItIsSkitty:
my fault, should be fixed now

PR #2125, opened 2025-03-03
Request for my tournament badge

llsebll:


Sendouc:
Looks too high res and not centered (vertically)

PR #2142, opened 2025-03-17
Update lfg.json

julufm:
Added full translation for French


julufm:
I put everything together

PR #2167, opened 2025-03-28
Add Stellaris Phantasm badges

andreaandradearagon:

## Badge adding checklist

- [X] The `.avif` while is static, not animated (generate it from the .png)
- [X] Rotation of animation is counterclockwise
- [X] Rotation is of correct speed (around 3 seconds for full round)
- [X] Badge is centered both vertically and horizontally
- [X] Badge is zoomed in enough (not too much blank space around it)

---

These three badges are for an upcoming tournament, each badge is meant to be given for the winners of each bracket.


Sendouc:
Hmm how was the texture generated? It looks like a different style than the existing badges.

andreaandradearagon:
You're totally right to point out that the texture style of my badge looks different from the existing ones. I just want to clarify that it was never my intention to go against the site’s guidelines or disrupt the overall style of the rest of badges.
I created the texture from scratch, resized to 128x128, imported it into picocad (to map the textures), and then brought it into the picocad web viewer with the original texture. Since it’s a star badge (and there are already quite a few of those) I thought using a custom palette might help make it stand out a bit more and feel more unique.
I actually tried the design without the custom texture at first, but it felt too plain and didn’t really capture the vibe I was aiming for. The final version just looked more complete and visually interesting with it.
Before deciding to go with a custom palette, I honestly wasn’t sure if it would be accepted. But I noticed some previous badges using pastel colors or slightly different styles, so I figured there might be some flexibility. One of the textures is also inspired by the black lotus badge.
I’ve already shared the badges on social media and people seem genuinely excited about it (there’s already a bit of hype going). That definitely encouraged me to submit it as-is, but of course, I’m totally open to making changes if needed to better fit the guidelines.
I completely understand the need for consistency, but with over 500 badges already, I think allowing the occasional custom palette could give artists more creative freedom and keep the badge system fresh and exciting for everyone. Also, sorry for bringing it up here (I just felt this was the best place to make sure my thoughts were heard). But if you think it’s better for it to be a feedback discussion, I’m totally fine with that!

Sendouc:
We can discuss it more on the feedback channel. I understand the argument for artistic freedom vs. consistency. But I also have to mind for the latter.

So I wanna also hear opinions from others.

PR #2180, opened 2025-04-06
add fracture bad for tournament i am going to host

S1MPLEsplatoon:

## Badge adding checklist

- [ ] The `.avif` file is static, not animated (generate it from the .png)
- [ ] Rotation of animation is counterclockwise
- [ ] Rotation is of correct speed (around 3 seconds for full round)
- [ ] Badge is centered both vertically and horizontally
- [ ] Badge is zoomed in enough (not too much blank space around it)

---


Sendouc:
The avif is not static. Also if you can i'd maybe zoom that in just slightly more. Will look better on the site.

PR #2229, opened 2025-04-27
Add badge for Tentatek Showdown

tariq-spl:
This PR adds the badge for the tournament Tentatek Showdown. 

**Creator:**
https://sendou.ink/u/tariq

**Notes:**
Ttek badge but faster


Sendouc:
The still pic should be from the front and not from an angle in this case and the badge is still wrong speed (about 3.3 secs now)

PR #2235, opened 2025-04-28
Add badge for LAN Championship Showdown (First)

thomatheto:
This PR adds the badge for the tournament LAN Championship Showdown (First). 

**Creator:**
https://sendou.ink/u/dotjpg


Sendouc:
Otherwise looks good but the badge is too fast (should be 3 seconds for a full spin)

PR #2251, opened 2025-05-06
Add badge for Crobattle Royale

mel0mania:
This PR adds the badge for the tournament Crobattle Royale. 

**Creator:**
https://sendou.ink/u/mel0mania


Sendouc:
Not sure what is happening here but there is just one badge added but 3 added in the config

Sendouc:
Okay ignore the above red explained on Discord it's some GitHub weirdness. But anyway same result because the speed is too fast & the rotation should be "upright" I don't know how else to explain it

PR #2258, opened 2025-05-08
Added Ical feed to calendar

### LINKED ISSUE #1476: Calendar as iCal feed
> One thing I'd love to see is making the event calendar (sendou.ink/calendar) available as an iCal feed or something similar that users can import into external calendar apps. (Google Calendar, etc.) You can keep the pretty webpage (it looks really nice ngl), just let us integrate it into our own calendars, Discord bots, and automated housemate reminder systems in a way that doesn't imply scraping.

> Came to ask for this pretty please. Did a search to see if it had already been asked and found the message I'm replying to, so I guess consider this a :plus1:. If this ever is implemented: setting the "Location" field to the url of the event on sendou.ink would be great. Tysm for your hard work!

kytoaa:
added iCal generation to the calendar, and a button to the calendar page to copy the ical link

closes #1476

I tested generated files with [the iCal validator](https://icalendar.org/validator.html), and by changing the link to an http localhost to connect it to a local calendar app


Sendouc:
Really awesome work 💯 🔥 Thank you for making the PR.

Couple things up for discussion/as a note:

1) Does using some library to do the feed generation e.g. https://www.npmjs.com/package/ics instead of raw string manipulation make sense? I'm thinking for example if we want to later add new fields say if we supported geo-data for LANs then we add on that. So wondering about the maintainability. But I don't feel strongly one way or another just wondering how you see it.

2) I got some work incoming here https://github.com/Sendouc/sendou.ink/tree/new-calendar and I can adapt your PR no problem.

3) Would it make sense to extract the ical code a calendar/core/ical.ts module? I'm asking because I'd like to build on top of this feature in the future. Mainly so that people can subscribe to their personal calendar (which would show all the events they are signed up for, TO'ing for and scrims) or team calendar (same but for teams). So I envision those would be different endpoints so I'd like to reuse as much of this code as possible when making those.

kytoaa:
I think it would definitely make sense to extract the ical code into a module, and with that it would probably also be good to have a more explicit type for events than `CalendarRepository.FindAllBetweenTwoTimestampsItem`

In my opinion currently a library probably wouldn't simplify it much, however, as you're thinking of expanding on it in the future though it would probably be good to use one.

I can start working on changing it to use a library tonight

Sendouc:
> I think it would definitely make sense to extract the ical code into a module, and with that it would probably also be good to have a more explicit type for events than `CalendarRepository.FindAllBetweenTwoTimestampsItem`
> 
> In my opinion currently a library probably wouldn't simplify it much, however, as you're thinking of expanding on it in the future though it would probably be good to use one.
> 
> I can start working on changing it to use a library tonight

Okay! But yeah completely up to you I think not using library also has its upsides so whatever you feel is the best. Extracting it into a module is the more important one of the changes I think.

kytoaa:
Moved the ical generation functions into `calendar/core/ical.ts` and switched to using https://www.npmjs.com/package/ics
its still passing my checks!

app/features/calendar/core/ical.ts
```diff
@@ -0,0 +1,71 @@
...
+		startTime,
+		endTime,
+		tagsToFilterBy: tagsFilter,
+		onlyTournaments: tournamentsFilter,
+	});
+	if (events.length === 0) return null;
+
+	const { error, value } = eventsAsICal(events);
+
+	if (error) return null;
```
Sendouc, line 38: Log error? I guess it should not happen. ```logger.error(`Error happened with construction the ical ${...}`)``` or something?

app/features/calendar/core/ical.ts
```diff
@@ -0,0 +1,71 @@
...
+		tagsFilter,
+		tournamentsFilter,
+	}: {
+		tagsFilter: Array<PersistedCalendarEventTag>;
+		tournamentsFilter: boolean;
+	} = { tagsFilter: [], tournamentsFilter: false },
+): Promise<string | null> {
+	const startTime = new Date();
+	const endTime = new Date(startTime);
+	// get all events over the next 3 weeks
```
Sendouc, line 21: Question: how does this work in ical? Why 3 weeks specifically? Or is this arbitrary

Sendouc (approved):
Looks great! Left a few comments.

I will think how it should integrate to my calendar changes. Maybe it can go in separately then I will rework it to my changes later 🤔 but I will think just a bit how to do it. Thank you again!

app/features/calendar/core/ical.ts
kytoaa: 3 weeks is arbitrary, I guess it might make more sense to make it a month or two

PR #2285, opened 2025-05-17
Update editorconfig

hfcRed:
Update editorconfig to align with codebase formatting. This should also fix the issue of git tracking every single line in the codebase sometimes. 
Biome also now uses editorconfig as a base for formatting.


Sendouc:
This was a really nice find!

PR #2287, opened 2025-05-17
Fix switch styling

### LINKED ISSUE #2161: SendouSwitch component broken styles
To reproduce log-in and navigate to settings page. You will notice if you make one of the labels long the Switch component looks a bit crazy.

Expected the Switch has min-width or some such so this can't happen.

Image

hfcRed:
Prevent switch from shrinking when label is too large. Just some flex weirdness, switched to grid instead.

Closes #2161

### DIFF
--- a/app/styles/elements.css
+++ b/app/styles/elements.css
@@ -137,7 +137,8 @@
 }
 
 .react-aria-Switch {
-	display: flex;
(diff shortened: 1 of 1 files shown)


Sendouc:
Ah interesting

PR #2319, opened 2025-05-21
Tranlation fr-EU update

julufm:
Added missing translations and corrected some translations for q.json and common.json files


julufm:
With pleasure, as usual!

PR #2330, opened 2025-05-25
Matckmaking UI

### LINKED ISSUE #2307: Clarify SendouQ team matchmaking UI
Swap visual importance of the "gap" and rank OR another rank. Anything else to make it more understandable?

See: https://discord.com/channels/299182152161951744/1339695498314911875/1339703524354818119

hfcRed:
This PR updates the UI of the groups displayed in Q to hopefully make it clearer to end users. 

Closes #2307 

Instead of a +- we now just display the two possible ranks

chrome_RZPRaMXjEi

### DIFF
--- a/app/features/sendouq/components/GroupCard.tsx
+++ b/app/features/sendouq/components/GroupCard.tsx
@@ -160,22 +160,26 @@ export function GroupCard({
 					</div>
 				) : null}
 				{group.tierRange?.range ? (
-					<div className="stack sm items-center">
-						<div className="q__group__tier-diff-text">
-							±{group.tierRange.diff}
+					<div className="stack items-center xs">
+						<div className="q__group__display-group-range font-bold">
+							<TierImage tier={group.tierRange.range[0]} width={80} />
+							{t("q:looking.range.or")}
+							<TierImage tier={group.tierRange.range[1]} width={80} />
+							<span className="text-lighter text-xs">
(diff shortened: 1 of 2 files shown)


hfcRed:
From my understanding the range was the primary source of confusion because people interpreted it as a full range, meaning if it said +-4 they thought they could also get +2 or -1. In this redesign, we ditch the range display and instead shift focus to the "or" which indicates only two possible outcomes.

Where would you put the range? Maybe we could also elaborate on this through text. For example we could put "One of two" above, so it would read like "One of two, Silver+ or Gold+". We could take it a step further and just slap an info popup with an explenation in there as well, but that feels a little clunky...

Maybe worth revisiting the thread to ask for opinions/ideas.

Sendouc:
https://github.com/Sendouc/sendou.ink/issues/2307#issuecomment-2895646556

I would just add the range like shown here. Then it describes a bit it's not just some two arbitrary ranks but the symmetry of it. But tbh still it's a bit confusing so a InfoPopover above the "or" might be needed to explain it.

Also the replay might need to be moved to make space for the small numbers (to the button text? so it displays "Challenge (Replay)")

Sendouc:
Oh yeah i like it on top!

I wonder if we could actually make the "or" a trigger for popover explaining it 🤔 kinda like same styling as here: https://github.com/Sendouc/sendou.ink/pull/2347 what do you think?

PR #2336, opened 2025-05-29
Add title to flags so that hovering shows country name

dangerous-beans:
- I never recognize the flags so I thought it'd be useful to be able to hover over it and see the name.
- I tried using the user's language for localization, instead of displaying directly in English, but I found it tricky to get the information.
- No dependencies introduced.
- I don't think this requires any testing other running `npm run dev` and going to a profile.

### DIFF
--- a/app/components/Flag.tsx
+++ b/app/components/Flag.tsx
@@ -13,6 +13,7 @@ export function Flag({
 				"twf-s": tiny,
 			})}
 			data-testid={`flag-${countryCode}`}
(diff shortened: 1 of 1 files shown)


dangerous-beans:
Okay this took an absurd amount of time for what it was. For some reason useTranslation() kept returning an empty object until... it spontaneously stopped doing so.

Thanks!

dangerous-beans:
Gonna fix the linter in a second

Sendouc:
Nice contribution!

PR #2340, opened 2025-05-30
Sort members on team page by role

dangerous-beans:
As the title says, sort members on team page by role, according to the order in the TEAM_MEMBER_ROLES array in team-constants.ts

This is something that has been annoying me and my team for a while, since the team page sometimes shows coaches and cheerleaders above team members.

My idea would be to eventually add the possibility to set the order in the "manage roster" page, but I believe this involves changes to the DB and it's beyond my abilities at this point.

PS: I have no idea how to do pull requests, I might be messing it up

### DIFF
--- a/app/components/Flag.tsx
+++ b/app/components/Flag.tsx
@@ -13,6 +13,7 @@ export function Flag({
 				"twf-s": tiny,
 			})}
 			data-testid={`flag-${countryCode}`}
+			title={new Intl.DisplayNames(["en"], { type: "region" }).of(countryCode)}
 		/>
 	);
 }
--- a/app/components/Flag.tsx
+++ b/app/components/Flag.tsx
@@ -15,7 +15,9 @@ export function Flag({
 				"twf-s": tiny,
 			})}
 			data-testid={`flag-${countryCode}`}
-			title={new Intl.DisplayNames([i18n.language], { type: "region" }).of(countryCode)}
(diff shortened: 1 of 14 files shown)


Sendouc (changes requested):
This should be done on the backend (TeamRepository). In addition for frontend code this kind of in-place mutation is not a good idea especially with stateful values. I like the feature idea though!

dangerous-beans:
Okay this works, but somehow the `.then()` doesn't seem idiomatic in this project. Am I right on this? If so, are there any resources I could ~skim~ carefully read to understand a better way to do this?

PS: Let's pretend I made a mistake when putting it in the frontend, it wasn't at all that I had no idea where I was putting it.

Sendouc:
> Okay this works, but somehow the .then() doesn't seem idiomatic in this project. Am I right on this? 

You are right.

> If so, are there any resources I could skim carefully read to understand a better way to do this?

I would normally start to understand how the existing code that needs to do this same thing has been implemented and try to match that :) It anyway is important to use the coding style of the project.

dangerous-beans:
Is this better? :D

dangerous-beans:
There might be a bug on this, I don't think it'll handle empty roles very well. I'll check when I'm home

dangerous-beans:
Not really a bug, but it was putting members with no roles at the beginning. Now they are at the end, where they rightfully belong

Sendouc:
Nice :)

PR #2347, opened 2025-05-31
Add a time component

### LINKED ISSUE #2334: Add copy for discord button for tournament times
basically the title, it should be pretty easy to implement (and i would be willing to implement it) I'm just not if this feature is wanted/necessary. The main benefit would be that telling people the time of a tournament on discord would be easier (no more manual conversion into other timezones for other people or putting it into a discord timestamp generator)

Phil-hacker:
Adds a Time component with copy to discord functionality

closes #2334 

Screenshot_20250531_144333

### DIFF
--- a/app/components/Time.tsx
+++ b/app/components/Time.tsx
@@ -1,12 +1,12 @@
-import { useTranslation } from "react-i18next";
+import { useRef, useState } from "react";
+import * as React from "react";
 import { Dialog } from "react-aria-components";
 import { Popover } from "react-aria-components";
-import { useRef, useState } from "react";
-import { ClipboardIcon } from "./icons/Clipboard";
-import * as React from "react";
+import { useTranslation } from "react-i18next";
+import { useCopyToClipboard } from "react-use";
 import { SendouButton } from "./elements/Button";
 import { CheckmarkIcon } from "./icons/Checkmark";
-import { useCopyToClipboard } from "react-use";
+import { ClipboardIcon } from "./icons/Clipboard";
 
 export default function Time({
 	time,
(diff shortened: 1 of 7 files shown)


Phil-hacker:
let me know if anything should be changed

app/components/Time.tsx
```diff
@@ -0,0 +1,82 @@
...
+import * as React from "react";
+import { Dialog } from "react-aria-components";
+import { Popover } from "react-aria-components";
+import { useTranslation } from "react-i18next";
+import { useCopyToClipboard } from "react-use";
+import { SendouButton } from "./elements/Button";
+import { CheckmarkIcon } from "./icons/Checkmark";
+import { ClipboardIcon } from "./icons/Clipboard";
+
+export default function Time({
```
Sendouc, line 11: Maybe a more specific name for this component? Like even `TimePopover`

app/components/Time.tsx
```diff
@@ -0,0 +1,82 @@
...
+		<div>
+			<div
+				ref={triggerRef}
+				className="dotted clickable"
+				onClick={() => {
+					setOpen(true);
+				}}
+			>
+				{time.toLocaleString(i18n.language, options)}
+			</div>
```
Sendouc, line 53: Would it be possible to use the `trigger` prop of Popover instead and use a styled button? I guess the downside with this solution is that it's not keyboard accessible (can't focus)

Sendouc (commented):
A few comments for discussion

Sendouc:
Just one last thing but seems like on the scrim page it has some unintended padding:

image

Phil-hacker:
it's fixed now

PR #2353, opened 2025-06-01
Brella launched damage calc

### LINKED ISSUE #2331: Update damage calculator to account for half damage on brella shield launch
After some testing, it does seem like launched brella shields take exactly half damage from any weapon compared to pre-launch, so this should be reflected in the damage calculator.

Should we display the extra rows with the shields having double HP or with the damage multipliers being halfed? The former would align with how Bubbler is displayed, while the latter is technically more accurate, but might look more confusing. Either way would result in the same HTD.

https://discord.com/channels/299182152161951744/1375666453197619271

hfcRed:
This PR adds launched variants for brellas to the damage calc, as well as a bubbler image variant for its weakpoints.

Closes #2331

Hard coded additional damage rate entry per brella, used the dash lines from the Runspeed Up ability to visualized launched mode.

chrome_sq0hrNKtDP
chrome_cf2KS1Tz1N


hfcRed:
As a side effect, this PR also updates the ``object-dmg.json`` as a whole, as I've noticed it is not fully up to date

Sendouc:
Nice work. I love the icons!

PR #2376, opened 2025-06-08
Add CSS rule `touch-action: manipulation` to ability selector

lzm0:
I was trying to analyze an Ink Saver only build on the [build analyzer page](https://sendou.ink/analyzer) on a phone. 
Tapping the ability button twice will zoom in the page instead of selecting the ability twice. This behavior is a bit strange and can be fixed with a [`touch-action`](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) CSS rule.

### DIFF
--- a/app/styles/common.css
+++ b/app/styles/common.css
@@ -1219,6 +1219,7 @@ abbr[title] {
 	flex-wrap: wrap;
 	justify-content: center;
 	gap: var(--s-1);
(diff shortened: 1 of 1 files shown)


Sendouc:
Wow yeah that's much better. Didn't know this kind of CSS rule exists, thank you!

PR #2384, opened 2025-06-09
Add badge for Splatoon Olympics

Salvifique:
This PR adds the badge for the tournament Splatoon Olympics. 

**Creator:**
https://sendou.ink/u/steven


Sendouc:
Should be more zoomed in and I think it might be preferable to have some outlines otherwise it looks a bit off when it's rotating.

PR #2398, opened 2025-06-15
Rules Clarification / Updates (+ quick info page update)

DreamyVeemo:
hiya sendou  i made this quick little mock up of an updated rules pages, i feel like the rules page dose not have exactly all the info and sometimes the wording can be a bit weird so BAM! updated rules lmk if you want any changes or are even considering the change at all :D

(btw changed the info page to reflect the $500 cash prize instead of $750 im pretty sure you said this in this https://bsky.app/profile/sendou.ink/post/3lrllpzafe22h tweet) 

😎

### DIFF
--- a/app/features/sendouq/routes/q.rules.tsx
+++ b/app/features/sendouq/routes/q.rules.tsx
@@ -33,16 +33,15 @@ export default function SendouqRules() {
 				</ol>
 				For the replay same weapons and gear must be used by both teams. The
 				team who fails to do so loses the map. If players disconnect from both
-				teams no replay will be used from either team&apos;s one replay
-				for the set. <br /> <br /> If the host disconnects otherwise known as
-				a "lobby crash" (where all players disconnect from the match) then the
-				team not hosting can either, replay (if the rules above apply) or must
-				make the team that was hosting the room take the loss for that game.
+				teams no replay will be used from either team&apos;s one replay for the
+				set. <br /> <br /> If the host disconnects otherwise known as a "lobby
+				crash" (where all players disconnect from the match) then the team not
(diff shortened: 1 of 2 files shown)


app/features/sendouq/routes/q.rules.tsx
```diff
@@ -110,13 +115,25 @@ export default function SendouqRules() {
...
 				It&apos;s not allowed to spawncamp the enemy without pushing the
 				objective provided it can&apos;t be considered a viable strategy taking
 				in account the game situation.
 			</div>
 
+			<h2 className="text-lg mt-4">Cheating / Mods</h2>
+			<div>
+				Cheating with any sort of modded hardware and or software is {" "}
+				<b>STRICTLY PROHIBITED</b>.
+			</div>
```
Sendouc, line 129:
```suggestion
				Using any sort of modded hardware and or software is not allowed.
			</div>
```

Sendouc: Feels out of place to have that kind of emphasis when nowhere else is.

app/features/sendouq/routes/q.rules.tsx
```diff
@@ -110,13 +115,25 @@ export default function SendouqRules() {
...
+
+			<h2 className="text-lg mt-4">Aliases</h2>
+			<div>
+				Aliases are allowed, Teams can ask for players to use their IGN.
+				If asked, all players must use the listed IGN on the user&apos;s profile.
+			</div>
```
Sendouc, line 135: I don't really see the point of this. Seems redundant as you can use the FC to check you played against the correct player already.

Sendouc:
Thanks good clarifications though

PR #2402, opened 2025-06-18
Badge Rename for my tournaments

DreamyVeemo:
changed name of badges beceasue we either 
1. dont run them anymore
2.  are reusing  them for something else

thanks for all your hard work sendou! :D

### DIFF
--- a/app/features/sendouq/components/MemberAdder.tsx
+++ b/app/features/sendouq/components/MemberAdder.tsx
@@ -75,7 +75,7 @@ export function MemberAdder({
 					/>
 					<SubmitButton
 						variant="outlined"
-						_action="ADD_MEMBER"
+						_action="ADD_TRUSTED"
 						isDisabled={!truster}
 						icon={<PlusIcon />}
 					/>


Sendouc:
Note that the badges have to be awards for Splatoon tournaments.

DreamyVeemo:
Yup, instead of adding new badges I'd like to reuse old ones that don't have a purpose anymore.

Sakura -> Special
We don't run Sakura edition anymore so I was gonna use it for special Editions of that tournament.

Teto -> MVP
We stopped running the Teto one because nobody really liked it and I plan on using this for the winning team of my tournament.

Brush Brawls -> Award
This tournament that it was planned for never ran so idk what to do with it low-key expect give it out for winning a brush related tournament.

5 Unique COE -> COE Special 
We don't run 5 different tournaments soooo what's the point 😭 I was gonna use this for one time tournaments that we run

Hope this added clarification!
:D

DreamyVeemo:
naurrrr :( 
I was gonna make a kahoot related to Splatoon and the sendiu.ink is that to far from Splatoon to be counted as a "Splatoon tournament"?

PR #2403, opened 2025-06-18
Increase tournament organization max name length

AKonicki26:
Increases the character limit on an organization's name from 32 -> 64 characters. 64 is chosen to follow the 2^n pattern from 32


Sendouc:
thanks!

Issue #2304, opened 2025-05-18
Refactor pure SQLite3 queries to Kysely


Sendouc:
![Image](https://github.com/user-attachments/assets/42d91a2f-afa5-497f-838e-ea43729d330f)

Essentially anything inside /queries folders should be refactored into Kysely queries. Can(/should) be done in smaller PR's

Code inside brackets-manager can be left as is for now.

Phil-hacker:
Current Progress

- [ ] art
- [ ] auth
- [ ] badges
- [ ] ban
- [ ] build-stats
- [ ] builds
- [ ] img-upload
- [ ] leaderboards
- [ ] mmr
- [ ] sendouq
- [ ] sendouq-match
- [ ] top-search
- [ ] tournament
- [ ] tournament-bracket
- [ ] tournament-subs
- [ ] vods (#2423)

Sendouc:
I'll do at least `builds` and `build-stats` in the experimental branch.

Sendouc:
🎉

PR #2423, opened 2025-06-25
Replace the old sql queries with kysely queries for the VODs feature

Phil-hacker:
First PR in a series of PRs fixing #2304


Phil-hacker:
sorry for the late response, life got in the way

i was able to replace
```typescript
fn.agg("json_group_array", [
	fn.agg("json_object", [
		val("username"),
		"User.username",
		val("discordId"),
		"User.discordId",
		val("discordAvatar"),
		"User.discordAvatar",
		val("customUrl"),
		"User.customUrl",
	]),
])
.as("players"),
```
with
```typescript
jsonArrayFrom(
	eb
		.selectFrom("User")
		.select([
			"User.username",
			"User.discordId",
			"User.discordAvatar",
			"User.customUrl",
		]),
).as("players"),
```
but `jsonArrayFrom` creates `{rowName: T}[]` instad of `T[]`, so i wasn't able to replace the other parts of the query yet

i could create a safe wrapper for it, but i'm not sure how to do it

Sendouc:
Is there a specific one you struggle with? Maybe I can point you to the right direction. But yeah sometimes it's just easiest to do a small mapping on the JS side.

app/utils/kysely.server.ts
```diff
@@ -33,5 +34,20 @@ export function unJsonify<T>(value: T) {
...
+		eb
+			.selectFrom("User")
+			.select([
+				"User.username",
+				"User.discordId",
+				"User.discordAvatar",
+				"User.customUrl",
+			]),
+	).as("players");
+}
```
Sendouc, line 50: Is this a generic function? I might just define it in the `VodRepository.ts` file. Also won't this grab every single User we have in the database (since it has no `where`) 🤔

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,258 @@
...
+	userId,
+	limit = VODS_PAGE_BATCH_SIZE,
+}: {
+	weapon?: MainWeaponId;
+	mode?: ModeShort;
+	stageId?: StageId;
+	type?: Tables["Video"]["type"];
+	userId?: number;
+	limit?: number;
+}): Promise<Array<ListVod>> {
```
Sendouc, line 41: It might be better to spin off findVodsByUserId as its on function from here. I know the existing code was not like this but I think with the better code reusability it might read better. Or what do you think?

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,258 @@
...
+	}
+	if (weapon) {
+		query = query.where((eb) =>
+			eb.or(
+				weaponIdToArrayWithAlts(weapon).map((weapon) => {
+					return eb("VideoMatchPlayer.weaponSplId", "=", weapon);
+				}),
+			),
+		);
+	}
```
Sendouc, line 84: Instead we can use the `IN` operator

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,258 @@
...
+	const result = query
+		.groupBy("Video.id")
+		.orderBy("Video.youtubeDate", "desc")
+		.limit(limit)
+		.execute();
+
+	const vods = (await result).map((value) => {
+		const { playerNames, players, ...vod } = value;
+		const playerNamesArray: string[] = playerNames as string[];
```
Sendouc, line 93: We could avoid bypassing the Typescript compiler if we use Kysely more typesafe way. This example is from TournamentRepository:

```ts
jsonArrayFrom(
				eb
					.selectFrom("TournamentStaff")
					.innerJoin("User", "TournamentStaff.userId", "User.id")
					.select([
						...COMMON_USER_FIELDS,
						userChatNameColor,
						"TournamentStaff.role",
					])
					.where("TournamentStaff.tournamentId", "=", id),
			).as("staff"),
```

benefit of writing it like this is that the result is automatically typesafe without any tricks.

Sendouc (changes requested):
Thank you for taking a stab at this! Left some comments. Please ask me if any questions.

app/utils/kysely.server.ts
Phil-hacker: it's used in a few other places, but i will move it into `VodRepository.ts` because it would need modifications for most other queries, if i see it used more in later PRs I will move it into this file again

Phil-hacker:
this should be done now, let me know if i should change anything

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,276 @@
...
+				.updateTable("UnvalidatedVideo")
+				.set(video)
+				.where("id", "=", args.id)
+				.executeTakeFirstOrThrow();
+			videoId = args.id;
+		} else {
+			const result = await trx
+				.insertInto("UnvalidatedVideo")
+				.values(video)
+				// as id is needed for sqlite, see https://sqlite.org/forum/forumpost/033daf0b32 or the docs for `returning`
```
Sendouc, line 227: This doesn't seem correct. See BuildRepository for example

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,276 @@
...
+		for (const match of args.matches) {
+			const video_match_result = await trx
+				.insertInto("VideoMatch")
+				.values({
+					videoId: videoId,
+					startsAt: hoursMinutesSecondsStringToSeconds(match.startsAt),
+					stageId: match.stageId,
+					mode: match.mode,
+				})
+				// as id is needed for sqlite, see comment above
```
Sendouc, line 241: here as well

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,276 @@
...
+			const result = await trx
+				.insertInto("UnvalidatedVideo")
+				.values(video)
+				// as id is needed for sqlite, see https://sqlite.org/forum/forumpost/033daf0b32 or the docs for `returning`
+				.returning("UnvalidatedVideo.id as id")
+				.executeTakeFirstOrThrow();
+			videoId = result.id;
+		}
+		for (const match of args.matches) {
+			const video_match_result = await trx
```
Sendouc, line 233: camelCase

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,276 @@
...
+			submitterUserId: args.submitterUserId,
+			validatedAt: args.isValidated
+				? dateToDatabaseTimestamp(new Date())
+				: null,
+		};
+		if (args.id) {
+			await trx
+				.deleteFrom("VideoMatch")
+				.where("videoId", "=", args.id)
+				.executeTakeFirstOrThrow();
```
Sendouc, line 215:
```suggestion
				.execute();
```

You are not returning any value

app/features/vods/VodRepository.server.ts
```diff
@@ -1,5 +1,276 @@
...
+
+			await trx
+				.updateTable("UnvalidatedVideo")
+				.set(video)
+				.where("id", "=", args.id)
+				.executeTakeFirstOrThrow();
```
Sendouc, line 221:
```suggestion
				.execute();
```

Phil-hacker:
this should be done now

Sendouc:
Thank you! Nice work.

PR #2440, opened 2025-07-03
Add i18n support format distance

### LINKED ISSUE #2418: Support i18n in `formatDistance`
`formatDistance` from date-fns is currently always English (used at least in notifications and scrims page).

It does support localization:

```ts
@example

// What is the distance between 1 August 2016 and 1 January 2015 in Esperanto?
import { eoLocale } from 'date-fns/locale/eo'
const result = formatDistance(new Date(2016, 7, 1), new Date(2015, 0, 1), {
  locale: eoLocale
})
//=> 'pli ol 1 jaro'
```

But how this should be done so it's convenient to use in this project. Some sort of custom hook that hooks it up with `useTranslation`? How to prevent loading unnecessary locales (or is it a problem if they are not too large? need to investigate).

GreenMarco:
Fixes #2418

This PR adds full i18n support for all formatDistance calls.

Main changes:
- Created a shared `formatDistanceWithI18n` function usable in both backend and frontend.
- Updated the changelog loader to pass the locale and removed caching for the changelog to prevent stale language issues.
- Modified backend functions that use formatDistance (e.g. in PlusSuggestionRepository.server.ts and Changelog.server.ts) to accept and use the current locale.
- Updated the `PostTime` component to use the `useFormatDistance` hook which gets the language from react-i18next.

How to test:
1. Change the language using the language selector.
2. Verify that relative dates in the changelog and posts update accordingly to the selected language.
3. Reload the page (F5 or Ctrl+F5) and confirm dates update correctly without showing cached content in a different language.

### DIFF
--- a/app/modules/in-game-lists/weapon-ids.ts
+++ b/app/modules/in-game-lists/weapon-ids.ts
@@ -26,7 +26,7 @@ export const weaponCategories = [
 	},
 	{
 		name: "BRUSHES",
-		weaponIds: [1100, 1101, 1110, 1111, 1112, 1122, 1115, 1121, 1120],
+		weaponIds: [1100, 1101, 1110, 1111, 1112, 1115, 1120, 1121, 1122],
 	},
 	{
 		name: "CHARGERS",


Sendouc:
Hey there thanks for the PR. Lint seems broken check `how-to.md` for more instructions.

Also I'm a bit concerned that this solution just pulls in a lot of extra stuff to the bundle most people wont use (the new locales). Somehow they should be lazy imported or something.

GreenMarco:
Lazy imported just work on backend? or there is an alternative to work on frontend?

GreenMarco:
Sorry, I'm new working on open sources codes

PR #2470, opened 2025-07-24
Add LD4 badges

Bouhtade:

## Badge adding checklist

- [x] The `.avif` file is static, not animated (generate it from the .png)
- [x] Rotation of animation is counterclockwise
- [x] Rotation is of correct speed (around 3 seconds for full round)
- [x] Badge is centered both vertically and horizontally
- [x] Badge is zoomed in enough (not too much blank space around it)

---

Files were generated using the tool, but I unfortunately couldn't get the pull request to be created from it, so I had to do it manually (hope that's not a problem).

This PR adds the badges for the french league LD4.


Sendouc:
Not centered and that base texture looks a bit too complex so it's of different style than existing badges.

Bouhtade:
For the centering most of the model is on the lower part of the badge but that's because the star on the top takes a lot of space, should I move it? And for the colors, yeah I'm using an unlocked color palette because I wanted the logo of the org to have it's actual color, would getting rid of the gradient on the trophy to use 2 flat colors for the base and the rest of the trophy be okay or should I also get rid of the gradient on the logo?

Bouhtade:
Yeah I think I get what you mean (except the grain idk where that comes from I didn't add it) But I just remembered that I left all the most recent files of the badge on my main computer and I'm away from home for like 2 weeks 💀
I would've made the stars unshaded and the model centered but I forgot to do that
Also idk what I was cooking with that gradient tbh it doesn't even make sense

PR #2475, opened 2025-07-29
Update i18n

hfcRed:
Fully updates the i18n to the proper format to cut down dev server startup time by almost 1000% cause I cant take it anymore😭

For further optimizations, things like gear and weapon names could be combined into an array and stored in a single key since they can only ever be loaded all at once regardless, which would cut down startup time considerably as well


Sendouc:
Frankly 1000% improvement is the least that is acceptable (nice one 🚀 )

PR #2478, opened 2025-08-01
Add player country info to teams in tourney API

soap-less:
Small change to `GET /api/tournament/{tournamentId}/teams` to expose country info. Helps in allowing TOs to pull statistics on which regions are playing in their tournaments as well as useful for pulling information for wikis like Liquipedia or Inkipedia should the community want to focus efforts in it down the line.

I didn't see any tests for the public API so apologies if there's a test I need to edit !!


Sendouc:
Here I think types are doing most of the heavy lifting for "tests".

PR #2498, opened 2025-08-20
Adds remaining fields for Sendou.ink Twitch Extension

### LINKED ISSUE #2492: Add Fields to GlobalTeamMembership and SendouqMatchPlayer
The Sendou.ink Twitch extension requires some extra fields to display all the necessary data. This requires adding fields to both `GlobalTeamMembership` and `SendouqMatchPlayer`

1. My extension calls the `/api/user/{userId|discordId}` endpoint when registering a user's Twitch account to get their associated Sendou.ink info. Here we take the team information but the current `GlobalTeamMembership` does not provide access to the url of the team's logo. Preferably, we should have
```
interface GlobalTeamMembership {
	/**
	 * Name of the global team.
	 *
	 * @example "Moonlight"
	 */
	name: string;
	/**
	 * Role of the user in the team.
	 */
	role: TeamMemberRole | null;
	/**
	 * URL for the global team page.
	 *
	 * @example "https://sendou.ink/t/moonlight"
	 */
	teamPageUrl: string;
    /**
     * URL for the global team logo.
     *
     */
    teamLogo: string
}
```

2. When loading a streamer's live Sendouq match, we require the following data: the members of each team with their associated Sendou.ink name, profile picture, url, and sendouq rank; and the maplist for the sendouq match. Regarding user data, we only have access to the user's Sendou.ink id and their Sendouq rank. We also already have access to the maplist. We still need each player's url, name, and profile picture. 

    Diving a little deeper, I noticed Sendou.ink does not always have the up-to-date profile picture for a user when calling `GetUserResponse`, this causes issues for my extension as Sendou.ink sometimes provides a stale profile picture link and so I have opted to grab profile pictures from Discord's API itself. However, to do this I need the user's discordId.

    There are two ways this can be solved. The first is as follows. When a live streamer enters a SendouQ match, we grab the match details, and then make 8 separate API calls to the `/api/user/{userId|discordId}` endpoint for each player in the match. This allows us to retrieve the data we need. The second solution is to change `SendouqMatchPlayer` to the following.

    ```
    type SendouqMatchPlayer = {
	    userId: number;
        discordId: string;
        /* User's Sendou.ink url(not to be confused with avatarUrl) */
        url: string;
        name: string;
	    /** User's at the start time of the match */
	    rank: SendouQRank | null;
    };
    ```
    This second method allows me to make a single API call to `/api/sendouq/match/{matchId}` for all the information I need.

Shak-codes:
Closes #2492 
- Adds `discordId` and `name` as fields for `SendouqMatchPlayer`
- Adds `avatarId` as field for `GlobalTeamMembership`


app/features/api-public/schema.ts
```diff
@@ -82,6 +82,8 @@ type SendouqMatchTeam = {
 
 type SendouqMatchPlayer = {
```
Sendouc, line 83: Instead of adding fields here you could just query the /user/{identifier} endpoint

app/features/api-public/routes/user.$identifier.ts
```diff
@@ -146,6 +152,7 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
 			name: team.name,
 			role: team.role,
 			teamPageUrl: `https://sendou.ink/t/${team.customUrl}`,
+			avatarId: team.avatarId,
```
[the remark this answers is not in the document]
Sendouc: Also I'm kind of having second thoughts about this endpoint. Maybe instead in this context it should contain the minimal fields: so role & team id. Then create a new endpoint for fetching team details based on ID and have the avatar, name & url there. This would be better aligned with how the API has otherwise been designed.

app/features/api-public/routes/user.$identifier.ts
```diff
@@ -74,11 +74,12 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
...
-						.select([
-							"Team.name",
-							"Team.customUrl",
-							"TeamMemberWithSecondary.role",
-						])
+						.leftJoin(
+							"UserSubmittedImage",
+							"UserSubmittedImage.id",
+							"Team.avatarImgId",
+						)
```
Sendouc, line 81: This join is a left over, right?

app/features/api-public/routes/team.$identifier.ts
```diff
@@ -0,0 +1,52 @@
...
+			.where("Team.id", "=", teamId)
+			.executeTakeFirst(),
+	);
+
+	const logoUrl = team.logoUrl ? userSubmittedImage(team.logoUrl) : null;
+
+	const result: GetTeamResponse = {
+		id: team.id,
+		name: team.name,
+		teamPageUrl: `https://sendou.ink/t/${team.customUrl ?? team.id}`,
```
Sendouc, line 47:
```suggestion
		teamPageUrl: `https://sendou.ink/t/${team.customUrl}`,
```

Not the same logic as user page

Sendouc (changes requested):
Few things to check. I'll still do some manual validation after but otherwise looking good!

PR #2501, opened 2025-08-23
 feat: make highlights save button / form errors sticky

961853266hyt:
- Move save button outside form to create sticky positioning
  - Add FormErrors to sticky area for better UX visibility
  - Implement u__highlights-container and u__highlights-sticky-button CSS classes
  - Ensure form errors are always visible when using sticky save button

截屏2025-08-23 下午12 57 57


Sendouc:
Nice one! Thank you

PR #2519, opened 2025-09-04
Stronghold article for BEEP

3dg3runner:


3dg3runner:
Finally figured out what the issue was, should be ready to merge unless there's something I'm missing.

PR #2589, opened 2025-10-19
badge: add badge for Salmon Workers

Rocko30:
This PR adds the badge for the tournament Salmon Workers. 

**Creator:**
https://sendou.ink/u/rocko_spl


Sendouc:
1. It must be done with another PR. But note this PR would not be accepted in the first place because it uses same display name for each badge.
2. There is no guarantee. Sometimes faster, sometimes slower. Depends what else I have on my plate.

PR #2666, opened 2025-12-18
Switzerland badge v2

Gaming32:
This PR adds badges for reaching 2000 SP and 2500 SP milestones for my tournament, Switzerland. The PR also updates the original Switzerland badge to match the style of these new badges.

Side note, what happened to the PR template?


Sendouc:
The flag should be more zoomed in.

> Side note, what happened to the PR template?
The badges one? Took it out as people should anyway add badges via the tool

PR #2680, opened 2025-12-29
Increase precision of `GetTournamentTeamsResponse.seedingPower`

Gaming32:
This removes the arbitrary precision limit of 2 decimal places for API use. This data doesn't really need to be truncated here as it's provided for machine consumption, not human display.


Gaming32:
I'm using the values for calculations to seed my own rating system.

Gaming32:
I start player power at `ord * 10 + 1500` and RD at `350 - |ord| * 3.75`. Doing math with rounded numbers is generally a pretty bad idea. Btw this is the only thing still keeping me on the `.data` "API".

Sendouc:
I see but I think for now I keep my opinion that the API already does what it should do and everything besides that is an implementation detail. Thank you for the PR but will be closing this for now.

PR #2759, opened 2026-01-20
Comp Analyzer

Sendouc:
Adds a new Comp Analyzer tool for analyzing weapon team compositions
 - Calculates damage combinations and kill times across selected weapons
 - Includes range visualization to compare weapon effective ranges
 - Sub/special weapon categorizations
 - Sub Defense Up/Ink Resistance Up calculations

### DIFF
--- a/app/features/comp-analyzer/comp-analyzer-hooks.ts
+++ b/app/features/comp-analyzer/comp-analyzer-hooks.ts
@@ -1,3 +1,4 @@
+import { MAX_AP } from "~/features/build-analyzer/analyzer-constants";
 import { useSearchParamStateEncoder } from "~/hooks/useSearchParamState";
 import type { MainWeaponId } from "~/modules/in-game-lists/types";
 import { mainWeaponIds } from "~/modules/in-game-lists/weapon-ids";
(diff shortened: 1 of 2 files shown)


Sendouc:
### Code review

Found 2 issues:

1. **Duplicate weapon selection allowed** - The `handleWeaponClick` function does not check if a weapon is already selected before adding it, allowing users to add the same weapon multiple times by clicking on an already-selected weapon in the grid.

https://github.com/sendou-ink/sendou.ink/blob/33ac22a3ffa18b39f24f2d7badcd92b8daa381a9/app/features/comp-analyzer/routes/comp-analyzer.tsx#L55-L67

2. **Undefined constant `WEAPON_CATEGORIES_WITHOUT_RANGE`** - This constant is referenced but never defined or imported in the file. This will cause a `ReferenceError` at runtime when the script executes. The file has `// @ts-nocheck` which hides this from TypeScript checking.

https://github.com/sendou-ink/sendou.ink/blob/33ac22a3ffa18b39f24f2d7badcd92b8daa381a9/scripts/create-analyzer-json.ts#L243-L251

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2781, opened 2026-01-28
Fix tentative tiers disappering

Sendouc:
"Hey, I was just looking at the calendar and I noticed upcoming editions of Teacup and Milkcup (https://sendou.ink/to/3184/register and https://sendou.ink/to/3193/register) don't have a ranking. Is there any reason why this might be? The previous editions of the tournaments have them."


Sendouc:
This approach is wrong https://github.com/sendou-ink/sendou.ink/issues/2783

PR #2846, opened 2026-03-01
fix: iCal link double slash

abysse-code:
## Summary

In file rendered by `https://sendou.ink/calendar.ics`, links are rendered with double slash, e.g. : `URL:https://sendou.ink//to/2979` instead of `URL:https://sendou.ink/to/2979`

## Routes changed

`"/calendar.ics", "features/calendar/routes/calendar.ics.tsx"`

## Notes

This is the only instance where a slash is appended to `SENDOU_INK_BASE_URL`


Sendouc:
Thank you!

PR #2961, opened 2026-04-09
Make <4v4 streams appear on front page, with a penalty

### LINKED ISSUE #2907: Stream ineligible tournaments are excluded even when a cast is active
Currently league divisions and tournaments <4v4 are excluded from the streams sidebar. I believe this should not happen if there is currently an active cast. For example, the current system will exclude LUTI Div X finals from the sidebar 😬.

Gaming32:
## Summary

This makes streams <4v4 appear on the front page again, but they're discounted based on their size. 3v3 is treated as one tier lower, and 1v1 is treated as 3 tiers lower.

## Routes changed

[`app/root.tsx`](https://github.com/sendou-ink/sendou.ink/blob/main/app/root.tsx)

## Notes

https://discord.com/channels/299182152161951744/784073459516964954/1489320064455938159

## Closes

#2907

### DIFF
--- a/app/features/core/streams/streams.server.ts
+++ b/app/features/core/streams/streams.server.ts
@@ -21,6 +21,7 @@ export type SidebarStream = {
 	subtitle: string;
 	startsAt: number;
 	tier: TournamentTierNumber | null;
+	membersPerTeam?: number
 	tentativeTier?: number;
 	peakXp?: number;
 	twitchUsername?: string;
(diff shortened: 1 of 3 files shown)


Sendouc:
Thanks for the PR! I did it slightly differently you can check it here: https://github.com/sendou-ink/sendou.ink/commit/0fbd851b91c3036b98ff8487c68fcfc241494594

PR #2968, opened 2026-04-12
Hide tournaments from Streams section if it has no streams

### LINKED ISSUE #2965: Check tournaments are actually live for sidebar
Currently might show even many tournaments as live with zero streams. Only consider a stream live if it has actually at least one cast or player stream

Gaming32:
Fixes #2965


Sendouc:
Thanks!

PR #2976, opened 2026-04-13
Correct higher/lower seed order in Match Map Info

### LINKED ISSUE #2975: Map bans using Higher Seed and/or Lower Seed in custom flow are attributed to the wrong team
Custom flow for testing:
Image

Match used for testing:
Image

Result after "eeltail haters" bans Eeltail Alley:
Image

Bans following the initial pick seem fine, however.

ximk:
## Summary

Fixes the order of `HIGHER_SEED` and `LOWER_SEED` to correctly display map bans affected by the "Higher Seed" and "Lower Seed" options during custom flow map selection.

## Closes

Closes #2975

### DIFF
--- a/app/features/tournament-bracket/components/MatchMapInfo.tsx
+++ b/app/features/tournament-bracket/components/MatchMapInfo.tsx
@@ -31,8 +31,8 @@ export function MatchMapInfo({ teams }: { teams: [number, number] }) {
 			teams,
 			results: data.results,
 			seeds: {
-				[teams[0]]: teamOne?.seed,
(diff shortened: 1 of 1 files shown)


ximk:
Converting to draft to fix the error again, as if the lower seeded team is Alpha, this no longer works.

ximk:
Reopening the PR since it should now function correctly!

Sendouc:
Thanks for the bug report and PR! We need a bit different fix here to have one source of truth for this so I went ahead and did that.

PR #3059, opened 2026-05-10
add equalizer badge

DreamyVeemo:
add badge for equalizer


DreamyVeemo:
badge order should be: red, blue, orange (1,2,3)

PR #3090, opened 2026-05-21
added super sea snails badge

RubyZym:
## Summary

adding super sea snails badge


Sendouc:
Use the tool https://github.com/sendou-ink/sendou.ink/blob/main/docs/badges.md

PR #3151, opened 2026-06-11
Replace TanStack/react-charts library (issue #2293)

### LINKED ISSUE #2293: Switch TanStack/react-charts to a mainained alternative
Now unmaintained as explained here: https://github.com/TanStack/react-charts

See also https://github.com/Sendouc/sendou.ink/issues/1806

lderuijter:
## Summary
Replaces the unmaintained `react-charts` (TanStack) library with `react-chartjs-2` and Chart.js. The custom tooltip is preserved with ability icons, colored dots, and value/suffix display. Both `linear` (ability points analyzer) and `localTime` (season SP history) axis modes are supported.

## Routes
- `/analyzer` — ability stat graphs
- `/u/<username>/seasons` — SP history chart

## Notes
- Removed `react-charts` dependency, added `react-chartjs-2`, `chart.js`, `chartjs-adapter-date-fns`, and `date-fns`
- Custom external tooltip renders React nodes as labels (ability icons) which wasn't possible with Chart.js built-in tooltips
- Grid colors are theme-aware (dark/light mode)
- Date formatting uses the existing `useDateTimeFormat` hook for locale consistency

## Testing
1. Go to `/analyzer`, click on the movement tab and then click on the lab icon
2. Hover over the chart, make sure the tooltip shows the ability icon, AP value, and suffix
3. For the `localTime` chart, go to `/u/<username>/seasons` on an account with 2+ days of season skill data and verify the SP history chart renders with correct date labels and tooltip

## Closes
#2293


package.json
```diff
@@ -56,6 +56,8 @@
 		"@tldraw/tldraw": "3.12.1",
 		"@zumer/snapdom": "2.12.8",
 		"better-sqlite3": "12.10.0",
+		"chart.js": "^4.5.1",
+		"chartjs-adapter-date-fns": "^3.0.0",
```
Sendouc: Could we do without this dependency? It seems not something I want to add due to lack of updates and low usage.

package.json
```diff
@@ -56,6 +56,8 @@
 		"@tldraw/tldraw": "3.12.1",
 		"@zumer/snapdom": "2.12.8",
 		"better-sqlite3": "12.10.0",
+		"chart.js": "^4.5.1",
```
Sendouc:
```suggestion
		"chart.js": "4.5.1",
```

package.json
```diff
@@ -80,7 +82,7 @@
 		"qrcode.react": "4.2.0",
 		"react": "19.2.7",
 		"react-aria-components": "1.18.0",
-		"react-charts": "3.0.0-beta.57",
+		"react-chartjs-2": "^5.3.1",
```
Sendouc:
```suggestion
		"react-chartjs-2": "5.3.1",
```

app/components/Chart.tsx
```diff
@@ -15,143 +34,167 @@ export default function Chart({
...
 	valueSuffix?: string;
 	xAxis: "linear" | "localTime";
 }) {
-	const theme = useTheme();
 	const isHydrated = useHydrated();
+
+	// Ref to the Chart.js instance, allows proper cleanup between renders to prevent "Canvas is already in use" errors
+	const chartRef = useRef<ChartType<"line"> | null>(null);
+
+	// Give each chart a unique id
```
Sendouc: Lets remove low value comments like this one and at least "Get the tooltip data" below

app/components/Chart.tsx
```diff
@@ -15,143 +34,167 @@ export default function Chart({
...
+		);
+	}, []);
 
-					return val;
-				},
-			},
-		}),
-		[scaleFormatter, xAxis],
-	);
+	// Get the grid color based on the current theme
```
Sendouc: question: Is styling Chart.js with just CSS not an option?

app/components/Chart.tsx
```diff
@@ -15,143 +34,167 @@ export default function Chart({
...
-			scaleType: xAxis,
-			shouldNice: false,
-			formatters: {
-				scale: (val: any) => {
-					if (val instanceof Date) {
-						return scaleFormatter.format(val);
-					}
+	// Get the line colors from CSS variables
+	const colors = React.useMemo(() => {
+		if (typeof document === "undefined") return ["", "", ""];
```
Sendouc: Use an object, this is confusing.

Sendouc (changes requested):
Hey! Thanks for the PR. Some things I noticed:

1) weird black line here
Image

2) line graph animation looks buggy (just disable the animation)

https://github.com/user-attachments/assets/e7a53b2b-d303-446c-922e-9051da979cc0

3) chart label color is changed from old version
4) There is a layout shift when the page loads
5) The chart feels too cramped with so many labels on seasons page

Feel free to ask for clarification either here or Discord if needed. Overall I think this looks solid just some polish to get it over the line 💪🏻

app/components/Chart.tsx
lderuijter: Yeah, sadly that's not an option because Chart.js renders onto a canvas element, and CSS cannot style anything drawn inside that canvas.

package.json
lderuijter: I removed it successfully.

app/components/Chart.tsx
lderuijter: Updated in last few commits.

package.json
lderuijter: Seems like removing this causes an error in the e2e test because the page needs the date adapter to work properly. I will add it back for now.

Sendouc: I think it is because we use `type: "time"` we _might_ get away with using "linear" as we already format those labels ourselves? But no worries if it seems too tricky. I can take a look after it's merged.

Sendouc:
image

seems like light mode is not supported currently. Lets make sure to use our css variables `vars.css`.

otherwise this is starting to look nice!

package.json
lderuijter: I removed the date adapter dependency successfully this time! I was able to just use `type: "linear"` when also passing indices and checking if its of the type Date later on.

Sendouc:
Thank you! This is great. Congratulations on the contribution as well :)

lderuijter:
> Thank you! This is great. Congratulations on the contribution as well :)

No problem! It was my pleasure :)

PR #3155, opened 2026-06-12
Updated Japanese analyzer.json

hyonemot:
Fixed some translations in the analyzer for Japanese. Used the Japanese Splatoon wiki as a reference


Sendouc:
Thank you!

PR #3163, opened 2026-06-15
Ingame name input

hfcRed:
Adds a new ingame name input with special character selection and stricter validation.
Character list mirrors Nintendo Switch keyboard characters 1:1 and validation blocks characters not allowed for ingame names.

The character selector is not in a popover because the text input would lose it's focus, the caret would not be movable anymore, and the text input can't be typed into until the popover is closed, which is bad UX imo. That does comes at the cost of the current solution causing a layout shift, which would not be desirable if we wanted to reuse the character selector for stuff like global search, tho that would need it's own implementation regardless, so I think this is fine as is.

15-15-27-53-chrome


Sendouc:
Awesome work as always!

PR #3164, opened 2026-06-15
Updated Japanese calendar

koreki750:
## Summary

Completed all remaining untranslated content.
Slightly changed wording for more natural flow and corrected mistranslations.

## Notes

Added "count.teams_one", "count.players_one", "count.teams_other", and "count.players_other" which were missing in the ja file compared to en.


Inkorest:
Sendou already added a script in a commit a few days ago to fix the issue where singular and plural forms are identical in zh/ja/ko.
For the two keys ending with '_one' and '_other', you just need to remove the suffixes and combine them into a single key, and you're good to go. ♪( ´▽｀)

Sendouc:
Yeah I should clarify the docs a bit on that! But that is fine I can combine those in a later commit.

PR #3172, opened 2026-06-17
chore: update "Country" in multi languages for better compliance

Inkorest:
## Summary

- Update "Country" label to "Country/Region" across the UI for better localization accuracy.

## Routes changed

None

## Notes

- I noticed this when I'm doing zh localization and updated it. Please let me know if you have any objections to this change! awa

## Closes

None


Sendouc:
Sure that makes sense. Just for any language not English it needs to be set to ""

Inkorest:
All done. Instead of leaving them blank, I've updated all the other languages with their correct translations. (๑´ڡ`๑)

PR #3183, opened 2026-06-21
Japanese translation progress

koreki750:
## Summary

Completed all remaining untranslated content.
Slightly changed wording for more natural flow and corrected mistranslations.

## Notes

I am not a developer and am not aware of how to phrase API, extensions, endpoints, and tokens. I used Japanese github as a reference but having someone who is familiar with the terminology check over it would be helpful.
Deleted translation for actions.neverMind since it was not used anywhere


Sendouc:
Thanks! We can always fix it later if needed.

PR #3231, opened 2026-07-16
Fix tab button styles being overriden by /plans

hfcRed:
When navigating to `/plans`, the route re-loads the `Button.module.css` file, moving it to the end of the document stylesheet array, which increases it's priority and overrides the tab buttons override styles until page reload.

16-11-30-40-chrome

Fixed by nesting the tab button styles which increases specificity, so it's override styles will always apply regardless of stylesheet order.

### DIFF
--- a/app/components/elements/Tabs.module.css
+++ b/app/components/elements/Tabs.module.css
@@ -19,27 +19,27 @@
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Sendouc:
Nice find!

PR #3233, opened 2026-07-16
Improve dev server performance on startup and navigation

hfcRed:
Finally got tilted enough at waiting 20 seconds for the dev server to spin up and 30 seconds for every navigation, so I spent the day looking for some performance improvements:

- Added `pnpm run dev:fast` which starts the server in 2 seconds vs 13 seconds after the initial `pnpm run dev`.
- React Compiler and Sentry plugins are now build only. Babels compiler was running on all 1000+ app modules on every navigation and reload for literally no reason?? Tradeoff is that components re-render more often in dev than prod, but I have not found any meaningful performance impact from this.
- Added `server.warmup` to pre-transform the main path, makes the first navigation to `localhost:5173` faster.
- Added `optimizeDeps.include` to stop the "new dependencies optimized" log thing mid-session, which forced a re-bundle and page reload on navigation. Literally no tradeoff other than the list has to be maintained, so if new dependencies trigger the above mentioned log, it should be added to the list. Saves 15-30 seconds per navigation and other cases which trigger re-bundles.
- Deferred seed import in the setup script, saves a few seconds on normal dev startup.

Video of a completely fresh install, first time running in dev, no cache or preload:

https://github.com/user-attachments/assets/60f42d63-a82f-47f2-94bb-6717fb5fe6e1

---

Some more considerations:

- Vites full bundle mode via `experimental.bundledDev` which would get us a lot of gains, but doesn't work with the current react router version, need to wait for official support.

- React router does warn us to upgrade Node versions, and from what I can tell there's no blocker to upgrade, so might be worth doing that as well for some performance gains.

- We could switch to tsgo via `@typescript/native-preview` but that seems to already be close to release so maybe we just wait it out.

I also did figure out that you can create a dev drive on Windows that you can move the project into which significantly speeds up the dev server, and that you can exclude the project from Windows Defender via
```shell
Add-MpPreference -ExclusionPath "C:\....\sendou.ink"
Add-MpPreference -ExclusionPath "C:\Users\user\AppData\Local\pnpm\store\v11"
```
which also speeds things up a lot, but is ofc risky since scripts installed via npm would no longer be checked by Windows Defender. Dev drive is pretty big tho ngl. Maybe worth documenting in README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Another key objective is to bridge the gap between casual and competitive player
 ### Prerequisites
 
 - [Git](https://git-scm.com/)
-- [Node.js v22](https://nodejs.org/en)
+- [Node.js v24](https://nodejs.org/en) (see [.nvmrc](./.nvmrc) for the exact version)
 - [pnpm](https://pnpm.io/installation)
 
 Optionally [nvm](https://github.com/nvm-sh/nvm) can be convenient for managing multiple Node.js installs
@@ -73,7 +73,7 @@ pnpm --version
 You should see something like:
 
 ```
-v22.13.0
+v24.18.0
 git version 2.39.5 (Apple Git-154)
 10.33.0
 ```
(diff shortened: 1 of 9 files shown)


Sendouc:
Wow that's indeed a lot smoother! I guess the biggest trade-off is a tiny bit less "prod-like" setup in dev but not too much so.

> React router does warn us to upgrade Node versions, and from what I can tell there's no blocker to upgrade, so might be worth doing that as well for some performance gains.

Where it says that?

> We could switch to tsgo via @typescript/native-preview but that seems to already be close to release so maybe we just wait it out.

I think it's already out as Typescript 7 so I think we can just do it ™️ 

also yes if there are some nice tips for smoother dev agreed should put to README

Again awesome job once more 🚀

hfcRed:
> Where it says that?

Okay that was just local to my machine whoops, I was still using the node version documented in the README. I updated it to point at the correct version now.

> I think it's already out as Typescript 7 so I think we can just do it ™️

I somehow completely missed the release?? Upgraded to TS7, finally. TSC went from 102 seconds -> 12 seconds on my machine, so a bit over a 8.5x speedup, server startup and editor type checking is also near instant now. Did have to mangle with the recursive form field names a bit, but ended up with a solution that's actually faster than what we had previously.

> also yes if there are some nice tips for smoother dev agreed should put to README

Added

Sendouc:
TS7 is just crazy fast. Awesome work the slowness was really driving me up the walls as well.

PR #3248, opened 2026-07-22
Fix broken YouTube thumbnails

### LINKED ISSUE #1304: Videos missing placeholder
image

Happens with videos that don't have `maxresdefault` thumbnail

hfcRed:
Strips url params and fragments from VOD video IDs before storing them so the `hqdefault.jpg` resolves properly

Closes #1304


Sendouc:
Very nice find! Thank you

PR #3259, opened 2026-07-28
Add share url button

hfcRed:
Adds a share button to the mobile nav menu so people using the PWA can still copy the current locations url

28-15-31-21-chrome


Sendouc:
I like this a lot!

PR #3301, opened 2026-08-04
Tourney custom themes

hfcRed:
Adds custom themes to established orgs and tournaments hosted by established orgs cause I lowkey forget that I wanted to add this earlier.

Custom themes are only available to established orgs and tournaments hosted by an established org.

04-15-29-13-chrome

04-15-29-33-chrome


Sendouc:
Orgs is really cool I like that.

For tournament I'm not fully sold on the idea. I fear that it will make the custom color pages quite a bit less special and might get a bit distracting for users as so many pages then have some different color scheme?

hfcRed:
It's been a highly requested feature. Maybe we can make just the info page a custom theme?

PR #3394, opened 2026-09-12
Trophy fixes

hfcRed:
Bumps the library version. Everything should be entirely non main thread blocking now, and a lot faster on Firefox specifically.

Also added the ability to define the model creator at upload time and change it later via the update trophy flow.

Unified some styles, made trophy loading a lot smoother overall. Fixed some WebGL bugs caused by React.

This should be the last PR before the feature flag is dropped (and XP trophies are hopefully added)


Sendouc:
Exciting!

PR #3398, opened 2026-09-16
Fix consent dialog covered by the mobile tab bar

### LINKED ISSUE #3397: Mobile: the tab bar covers the consent dialog's buttons
### What happens

On my phone the consent dialog that comes with the ads is cut off by the bottom tab bar: the primary button ("AGREE") sits behind it and can't be tapped at all. It isn't only painted over — a hit test at the button's centre returns the tab bar, so taps never reach the button.

I hit it on my iPhone in Safari and then reproduced it in WebKit with an iPhone 15 profile, and in Chromium at 375px wide, so it isn't iOS specific — iOS is just worse for it because of `env(safe-area-inset-bottom)`.

### Why it happens

`#mobile-nav` is `popover="manual"` and gets `root.showPopover()` on hydration (`app/components/MobileNav.tsx:99-101`), so the tab bar lives in the top layer. The consent dialog is drawn by the consent platform inside `fuse.js` into `#qc-cmp2-container`, the first child of `<body>`, as an ordinary element (`position: fixed`, `z-index: 2147483647`).

The top layer always paints above non-top-layer content, whatever the z-index, so the bar wins.

With the dialog up and the window under 600px wide:

```js
const button = [...document.querySelectorAll("#qc-cmp2-container button")].at(-1);
const { left, top, width, height } = button.getBoundingClientRect();

document.elementFromPoint(left + width / 2, top + height / 2);
// → <nav class="_tabBar_…">    the tab bar, not the button

document.getElementById("mobile-nav").hidePopover();   // leaves the top layer, stays rendered
document.elementFromPoint(left + width / 2, top + height / 2);
// → <span>AGREE</span>
```

### Since when

Before #3381 (2e26e1fe8) the bar was a plain `position: fixed; z-index: 19` element — `git show 2e26e1fe8^:app/components/MobileNav.module.css` — so the consent dialog painted over it normally. The top layer came in with that migration, and its changelog entry lists "Measures against ads covering the bottom navigation on mobile"; the consent dialog gets caught by the same measure.

### Reproducing

- viewport under 600px. At 600px and up `.mobileNav` is `display: none` and the bug disappears, so rotating the phone to landscape or widening a desktop window hides it
- an EEA visitor with no stored consent. Afterwards the footer's "Privacy" button (`data-fuse-privacy-tool`) re-opens the same dialog, covered in the same way
- ads on: `VITE_FUSE_ENABLED` defaults to false and isn't in `.env.example`, so it can't be reproduced against a local dev server or in the e2e run

### Impact

A first-time visitor on a phone in the EEA can't press the primary button of the consent dialog. "MORE OPTIONS" is still tappable, so there is a way through, but the one-tap path out isn't.

### Possible fix

Have the bar leave the top layer while the dialog is on screen and go back in once it's gone: the bar stays exactly where it is (its own `display: block` keeps it rendered outside the top layer), the dialog paints over it like it did before #3381, and ads still can't cover the bar the rest of the time.

No third-party class names needed — the consent platform implements the IAB TCF API. `__tcfapi('addEventListener', 2, …)` fires on every surfacing of the dialog (the privacy button included; a listener registered while it is already up is called immediately), and `__tcfapi('ping', 2, …)` answers `displayStatus: 'visible' | 'hidden'`, which also covers the dialog being closed without a choice, since TCF has no event for that.

---

Since CONTRIBUTING brings it up: I'm a user of the site, I ran into this on my own phone, and I went through the investigation and wrote the patch with an AI assistant helping me. Everything above is something I checked in a browser rather than guessed, and I'm happy to go through any of it. If you'd rather just have the report and fix it yourself, that's completely fine — feel free to close the PR.

PedroFlores199:
## Summary

On mobile the bottom tab bar covers the consent dialog that comes with the ads, so its primary button can't be tapped — a hit test over the button returns the tab bar, not the button. I ran into it on my iPhone in Safari, but it is not iOS specific: it reproduces in any browser under 600px wide, iOS is just worse for it because of `env(safe-area-inset-bottom)`.

`#mobile-nav` is a manual popover shown on hydration (`app/components/MobileNav.tsx:99-101`), which puts the bar in the top layer, and the top layer paints above every ordinary element whatever its z-index. The consent platform draws its dialog into `#qc-cmp2-container` as an ordinary element, so it always loses.

This lifts the dialog above the bar instead, the way `.publift-widget-sticky_footer-container` is already lifted a few lines above it in the same file. The platform's outer box is a full screen `position: fixed` with the dialog anchored to its bottom, so shrinking that box by the bar's height moves the buttons up. Nothing about the bar, the top layer or the panels changes.

| Before | After |
| --- | --- |
|  |  |

iOS Safari (iPhone 17 Pro Simulator, iOS 26.5). The site is running locally there with a stand-in for the dialog, since the consent platform only serves the real domain — the real thing, on sendou.ink in the same Safari, is in #3397.

## Routes changed

None specifically: it is the layout's mobile tab bar, so any page under 600px wide.

## Notes

- Measured against the live dialog in WebKit with an iPhone 15 profile: the platform's box goes from 812 to 757 px tall (exactly `--layout-nav-height`), the primary button becomes hit-testable, and the tab bar stays fully usable next to it.
- No `!important` needed: `#qc-cmp2-container .qc-cmp-cleanslate` beats the platform's own `.css-xxxxxxx.css-xxxxxxx` on specificity, and the rule sits with the other third-party rules outside `@layer base`, so an unlayered vendor stylesheet doesn't win over it either. `height: auto` is needed because their box sets `height: 100%`, which would otherwise beat `bottom`.
- The same box holds the dialog's second layer ("more options") and the dialog re-opened from the footer's privacy button, so both clear the bar too.
- The test renders the bar and a stand-in for the platform's box (an ordinary fixed element at `z-index: 2147483647` with a button at its bottom edge) and asserts with `elementFromPoint` that the button, not the bar, is the topmost thing there. It fails without the rule.
- It is pinned to the platform's class names, like the sticky footer rule above it. If they rename those, the rule stops matching and we are back to today's behaviour, never worse.
- The alternative I tried first was to have the bar leave the top layer while the dialog is up, reading `__tcfapi`'s `displayStatus` to know when. It works, but it is around 120 lines of listener, poll and cleanup for the same result, and it leaves the bar under the dialog rather than beside it. Happy to switch to that if you would rather not depend on their markup.

## Closes

#3397


Sendouc:
Thanks for looking into it and a fix!

PR #3401, opened 2026-09-19
SQ Match confirmation changes

hfcRed:
Adds cooldowns to game and match result buttons, makes it more clear when you're about to confirm a win or a loss, add double tap for loss confirmation, add button for telling the other team that the score is wrong


Sendouc:
Okay now it's definitely user error if wrong scores

PR #3407, opened 2026-09-21
Add scroll lock, replace CSS anchors with floating layer, and various fixes

hfcRed:
## Floating Layer

Replaces CSS anchors cause they kinda suck. Popovers now anchor properly, resize to take all available height, take into account top nav, mobile nav, and mobile keyboard for height calculations, are hidden when the anchor leaves the viewport (taking into account top nav, mobile nav), and the anchor is scrolled into view if the mobile keyboard were to cover it on open

https://github.com/user-attachments/assets/e2099bc4-f7bf-410b-b4e9-787f4d0ac29b

## Dialog Changes

Dialogs now properly scroll lock, have unified styles, and take into account the mobile keyboard for height calculations

https://github.com/user-attachments/assets/5e038e76-3a6e-4af3-a959-74942a841ce9

## Scroll Lock

Does what it says, hopefully for real this time if Apple actually wants to play nice. Please do test because I have no Apple devices. Added `--scrollbar-width` which fixed elements like toasts or the mobile nav can use for right padding to prevent a shift on scroll lock. 

https://github.com/user-attachments/assets/65c608e3-7de8-44d0-8932-57808fe880f6

### Top nav menus now carry over their open/closed state when hovering between them

https://github.com/user-attachments/assets/9c2bf7bf-7e28-451d-8929-a93369784916

### Click events on backdrops can no longer leak onto elements below (video of before)  

https://github.com/user-attachments/assets/44301300-da8f-4dc8-b162-0c60fa995539

### When logged out the search label now properly hides
20-19-00-48-chrome
21-10-11-11-chrome

### Mobile modals can now actually use all available height
21-10-14-06-chrome
21-10-14-48-chrome

## Stuff

- Added docs for overlays
- Updated README cause it was still saying we use React Aria Components


Sendouc:
Huge

